PR #1, opened 2024-06-26
Move vm-testing related files to separate directory, temporarily disable tests

bkabrda:
This PR is a little cleanup of the repository; it moves the VM-testing related files to a separate directory (to be fixed later) and temporarily disables the broken Github Action that uses these (it's been broken before moving the files around)


fghanmi (approved):
LGTM

PR #4, opened 2024-07-01
Fix idempotency of podman tasks and conditions for pod restarts

bkabrda:
This change addresses some issues around the podman logic:

* Ensures we restart podman pods under the right conditions (when config, secret or service file change)
* Implements proper idempotency for the related tasks
* Makes us depend on podman collection >= 1.15.0 which seems to have bunch of idempotency fixes


sabre1041 (approved):
LGTM

PR #5, opened 2024-07-03
[SECURESIGN-1163] | Check if component is enabled before pulling image

tommyd450:
Very much a noob PR. Let me know how I can improve this.

When I tested it skipped over pulling trillian and failed due to not having anything configured?

### DIFF
--- a/roles/tas_single_node/tasks/podman.yml
+++ b/roles/tas_single_node/tasks/podman.yml
@@ -17,9 +17,25 @@
   register: podman_login_result
   changed_when: '"Already logged in" not in podman_login_result'
 
+
+
 - name: Pull all images
   containers.podman.podman_image:
     name: "{{ item }}"
+  vars:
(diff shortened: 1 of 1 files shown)


roles/tas_single_node/tasks/podman.yml
```diff
@@ -32,6 +46,9 @@
     - "{{ tas_single_node_netcat_image }}"
     - "{{ tas_single_node_nginx_image }}"
     - "{{ tas_single_node_curl_image }}"
+  loop_control:
+    index_var: my_idx
```
bkabrda, line 50: Let's use `idx` instead of `my_idx`

bkabrda (changes requested):
Thanks for the PR. I only have one minor suggestion to fix, otherwise it looks good. Please address the comment and then I can approve.

bkabrda (approved):

PR #9, opened 2024-07-12
Avoid requiring resolution of base_hostname for applying tas_single_node

bkabrda:
We can fetch the rekor pubkey locally on the managed node. This is the only part that required the ability to resolve base_hostname when applying the role. With this fix, base_hostname no longer needs to resolve on the control node to apply the tas_single_node role.

Note: because we do this locally on the managed node and the pubkey doesn't go over any network, we can also use http instead of https.


fghanmi (approved):
lgtm

PR #10, opened 2024-07-12
[SECURESIGN-1113] | Necessary Changes to add and Have Rekor-Search-Ui Running

tommyd450:
Adds rekor-search-ui to the ansible stack,

Changes

- Rekor-Server now runs on port 3001 as Rekor Search requires port 3000 but Rekor-Server.
- Adds Rekor Search Ui

Note
Test script now works.
To have Rekor Search Ui be able to reach Rekor ensure that you downloaded and configured the certs locally as Search Ui runs the query to it on the clients browser.


roles/tas_single_node/tasks/podman.yml
```diff
@@ -80,3 +82,6 @@
 - name: Configure/Deploy TSA
   ansible.builtin.include_tasks: podman/tsa.yml
   when: tas_single_node_tsa_enabled | bool
+- name: Configure/Deploy RekorSearch
```
bkabrda, line 85: Nit: let's add newlines between individual tasks.

roles/tas_single_node/tasks/podman/nginx.yml
```diff
@@ -37,6 +39,14 @@
       | b64decode + (remote_ingress_certificates.results | selectattr('source', 'equalto', tas_single_node_remote_ca) | list | first).content
       | b64decode }}
 
+- name: Base64 encode the rekor ingress
+  ansible.builtin.set_fact:
+    rekor_search_ingress_base64: >-
+      {{ (remote_ingress_certificates.results | selectattr('source', 'equalto', tas_single_node_certs_dir + '/ingress-rekor-search.pem') | list | first).content
+      | b64decode + (remote_ingress_certificates.results | selectattr('source', 'equalto', tas_single_node_remote_ca) | list | first).content
+      | b64decode }}
+
```
bkabrda, line 48: Another nit: only one line between the tasks, unless there's a reason to create a larger visual space for some reason.

roles/tas_single_node/templates/manifests/rekor-search-ui/rekor-search-ui.yml
```diff
@@ -0,0 +1,39 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: rekor-search-ui
+  namespace: rekor-search-ui
+  labels:
+    app.kubernetes.io/component: search
+    app.kubernetes.io/instance: scaffold
+    app.kubernetes.io/managed-by: Helm
```
bkabrda, line 9: I don't think `managed-by: Helm` is appropriate in this file, let's just remove it.

roles/tas_single_node/defaults/main.yml
```diff
@@ -99,7 +99,7 @@ tas_single_node_trillian_secret: "{{ tas_single_node_kube_configmap_dir }}/trill
 tas_single_node_systemd_directory: /etc/systemd/system
 
 tas_single_node_setup_host_dns: true
-tas_single_node_base_hostname: ""
+tas_single_node_base_hostname: "myrhtas"
```
bkabrda, line 102: Let's not change this value. We want user to provide this and we want to fail if it is not provided.

bkabrda (changes requested):
Hi :wave: this looks very good. I left couple minor comments inline to be addressed. Let me know if you still have issues with running the test script, we need to make sure that is fixed and that you can run it before we merge.

roles/tas_single_node/tasks/podman/nginx.yml
tommyd450: I believe this was an issue pointed out by the linitng/sanity checks.

bkabrda (approved):

PR #11, opened 2024-07-12
[SECURESIGN-1219] Adding sanity testing to collection

SequeI:


roles/tas_single_node/templates/manifests/fulcio/fulcio-server.yaml
```diff
@@ -33,12 +33,12 @@ spec:
             - --fileca-cert
             - /var/run/fulcio-secrets/cert.pem
             - --fileca-key-passwd
-            - {{ tas_single_node_fulcio_ca_passphrase  }}
-            - --ct-log-url=http://ctlog-pod:6962/{{ tas_single_node_ct_logprefix }}
+            - "{{ tas_single_node_fulcio_ca_passphrase }}"
+            - --ct-log-url=http://ctlog-pod:6962/"{{ tas_single_node_ct_logprefix }}"
```
bkabrda, line 37: I think it's better to put the quotes around the whole string, not just part of it - I don't think I've ever seen usage of quotes mid-string and I'd be worried this is fragile and could behave differently on different versions of Ansible/Jinja2

bkabrda (changes requested):
Thanks for the PR, I left one comment inline and I would also like to ask you to add a note in the README.md file about running sanity testing locally - there's already a note about running ansible-lint locally, so let's add the sanity testing notes there. When these issues are fixed, I'll aprove.

roles/tas_single_node/templates/manifests/fulcio/fulcio-server.yaml
SequeI: Change made

PR #12, opened 2024-07-15
[SECURESIGN-1174] Add Timestamp Authority

fghanmi:
Add Timestamp Authority server on port 3002.
Updated vm_testing/test.sh to use it instead of Rekor TSA.

## Testing command:
```
./vm_testing/test.sh
```
Expected Result: `...Verified OK`

### DIFF
--- a/roles/tas_single_node/tasks/certificates.yml
+++ b/roles/tas_single_node/tasks/certificates.yml
@@ -64,6 +64,7 @@
     - fulcio
     - rekor
     - tuf
+    - tsa
   loop_control:
     loop_var: cert_name
   vars:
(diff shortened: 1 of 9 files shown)


roles/tas_single_node/tasks/certificates.yml
```diff
@@ -136,6 +137,102 @@
...
+      register: tsa_private_key_stat
+
+    - name: Create TSA root Private Key
+      community.crypto.openssl_privatekey:
+        path: "{{ tas_single_node_local_tsa_private_key }}"
+        curve: secp256r1
+        type: ECC
+        passphrase: "{{ tas_single_node_tsa_ca_passphrase }}"
+        cipher: auto
+      when: (tsa_private_key_stat.stat.islnk is not defined) and (tsa_chain_stat.stat.islnk is not defined) and tas_single_node_signer_type == 'file'
```
bkabrda, line 161: I could be wrong here, but I think you can just use `force: false` (see the [module doc](https://docs.ansible.com/ansible/latest/collections/community/crypto/openssl_privatekey_module.html)) to specify that the private key shouldn't be overwritten if it already exists (`false` is also the default, which is why the existing code doesn't seem to address this case at all - it's addressed implicitly).

roles/tas_single_node/tasks/podman/tsa.yml
```diff
@@ -0,0 +1,73 @@
...
+  register: tsa_server_yaml
+
+- name: Deploy TSA Pod
+  ansible.builtin.include_tasks: podman/install_manifest.yml
+  vars:
+    podman_spec:
+      state: started
+      systemd_file: tsa
+      network: "{{ tas_single_node_podman_network }}"
+      kube_file_content: "{{ tsa_server_yaml.content | b64decode | from_yaml }}"
```
bkabrda, line 71: For the other services, we currently use e.g. `"{{ lookup('ansible.builtin.template', 'manifests/nginx/nginx.yaml') | from_yaml }}"` without having to explicitly render the file on drive. Is there a reason why this wouldn't work here?

bkabrda (changes requested):
Hi, thanks for the PR, I think this looks very solid. I left few comments inline and I would also like to ask you to modify the test script under the `vm-testing/` directory to test the timestamp server as well.

roles/tas_single_node/tasks/podman/tsa.yml
fghanmi: In the other services, we don't use j2 templates as we are trying to do for TSA (we need conditionals) 
That's why here, we are rendering the j2 template into a yaml before using it.

bkabrda: But isn't that what [template lookup](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/template_lookup.html) does? Sorry if I'm missing something, but I think we should be able to do here what other services do (I haven't tried this myself, so I could be wrong though)

roles/tas_single_node/tasks/certificates.yml
fghanmi: I think it's better to use the same way as the rest of the certificates:
example: 
```
  when: >
    (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_fulcio_private_key) | list | length) == 0
```
(I updated it accordingly)

roles/tas_single_node/tasks/podman/tsa.yml
fghanmi: you're correct! Thanks!

roles/tas_single_node/tasks/podman.yml
```diff
@@ -44,33 +45,38 @@
...
     - "{{ tas_single_node_nginx_image }}"
+    - "{{ tas_single_node_tsa_image }}"
   loop_control:
     index_var: idx
   when: enabled_vals[idx] | bool
 
-- name: Configure/Deploy nginx
-  ansible.builtin.include_tasks: podman/nginx.yml
+# - name: Configure/Deploy nginx
+#   ansible.builtin.include_tasks: podman/nginx.yml
```
bkabrda, line 54: I think you commented out a lot of lines in this file for testing but forgot to uncomment?

bkabrda (approved):

PR #13, opened 2024-07-17
Fix syntax of ctlog url passed to fulcio server

bkabrda:
The quotes weren't recognized as YAML element, but as part of the string and passed verbatim to the fulcio server, which made it fail to communicate with ctlog server.


tommyd450 (approved):

tommyd450:
lgtm

PR #14, opened 2024-07-18
[SECURESIGN-1116] Add option to use external customer provisioned Redis

fghanmi:
### Description
- Add option to use customer provisioned Redis:
To do that:
```
tas_single_node_rekor_redis:
  database_deploy: false
  redis:
    address: "address"
    port: 6379
    password: "password"
```

--- 
- This option is already enabled for Trillian DB. User can disable the creation of the trillian DB using `tas_single_node_trillian.database_deploy ` set to false and provide `mysql` attributes:
```
tas_single_node_trillian:
  mysql:
    database_deploy: false
    user: mysql
    rootPassword: rootpassword
    password: password
    database: trillian
    host: trillian-mysql-pod
    port: 3306
```

### DIFF
--- a/roles/tas_single_node/defaults/main.yml
+++ b/roles/tas_single_node/defaults/main.yml
@@ -19,7 +19,14 @@ tas_single_node_skip_os_install: false
 
 tas_single_node_rekor_templates:
   - manifests/rekor/redis-server.yaml
-  - manifests/rekor/rekor-server.yaml
+  - manifests/rekor/rekor-server.j2
+
+tas_single_node_rekor_redis:
+  database_deploy: true
+  redis:
+    address: "rekor-redis"
+    port: 6379
+    password: ""
 
 tas_single_node_cockpit_enabled: true
 tas_single_node_ctlog_enabled: true
--- a/roles/tas_single_node/defaults/main.yml
+++ b/roles/tas_single_node/defaults/main.yml
@@ -18,15 +18,15 @@ tas_single_node_cockpit:
 tas_single_node_skip_os_install: false
 
 tas_single_node_rekor_templates:
-  - manifests/rekor/redis-server.yaml
+  - manifests/rekor/redis-server.j2
   - manifests/rekor/rekor-server.j2
 
 tas_single_node_rekor_redis:
   database_deploy: true
   redis:
-    address: "rekor-redis"
+    address: rekor-redis-pod
     port: 6379
-    password: ""
+    password: password
 
 tas_single_node_cockpit_enabled: true
 tas_single_node_ctlog_enabled: true
--- a/roles/tas_single_node/tasks/podman.yml
+++ b/roles/tas_single_node/tasks/podman.yml
@@ -17,30 +17,6 @@
   register: podman_login_result
   changed_when: '"Already logged in" not in podman_login_result'
 
-- name: Define images base list
-  ansible.builtin.set_fact:
-    images_list: [
-      "{{ tas_single_node_fulcio_server_image }}",
-      "{{ tas_single_node_trillian_log_server_image }}",
-      "{{ tas_single_node_trillian_logsigner_image }}",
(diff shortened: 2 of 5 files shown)


roles/tas_single_node/defaults/main.yml
```diff
@@ -19,7 +19,14 @@ tas_single_node_skip_os_install: false
 
 tas_single_node_rekor_templates:
   - manifests/rekor/redis-server.yaml
-  - manifests/rekor/rekor-server.yaml
+  - manifests/rekor/rekor-server.j2
+
+tas_single_node_rekor_redis:
+  database_deploy: true
+  redis:
+    address: "rekor-redis"
```
bkabrda, line 27: I'm wondering why this isn't `rekor-redis-pod`. I understand that this is the value that was used before and you just extracted it to this variable, but I'm struggling to understand why in other places we use `servicename-pod` syntax and here we just use `servicename`. Are both possible? Could you please look into this so that we understand? Thanks!

roles/tas_single_node/tasks/podman/rekor.yml
```diff
@@ -25,6 +25,7 @@
       systemd_file: redis
       network: "{{ tas_single_node_podman_network }}"
       kube_file_content: "{{ lookup('ansible.builtin.template', 'manifests/rekor/redis-server.yaml') | from_yaml }}"
+  when: tas_single_node_rekor_redis.database_deploy
```
bkabrda, line 28: We should also use this for conditionally (not) pulling the image in `tasks/podman.yml`. (If I'm looking right, we should also improve the condition for prepulling the trillian_db image with `tas_single_node_trillian.database_deploy` - could you please address that as well?)
Thanks!

roles/tas_single_node/defaults/main.yml
```diff
@@ -19,7 +19,14 @@ tas_single_node_skip_os_install: false
...
+    port: 6379
+    password: ""
```
bkabrda, line 29: Let's provide a default password for Ansible-managed redis while we're at it, just to make sure we have something (and be consistent with the MariaDB defaults)

roles/tas_single_node/defaults/main.yml
fghanmi: you're right! I'm wondering how it was "working" 
image

bkabrda: Maybe it didn't, but rekor can run without redis? Maybe it only uses it for caching, but can actually work (slower) without it?

fghanmi: yes, it's used for caching, it wasn't detected with the actual tests, since we don't test all rekor functionalities (and performance that depends on caching)

roles/tas_single_node/tasks/podman.yml
```diff
@@ -34,18 +58,7 @@
...
-    - "{{ tas_single_node_trillian_logsigner_image }}"
-    - "{{ tas_single_node_rekor_image }}"
-    - "{{ tas_single_node_ct_server_image }}"
-    - "{{ tas_single_node_redis_image }}"
-    - "{{ tas_single_node_trillian_db_image }}"
-    - "{{ tas_single_node_tuf_image }}"
-    - "{{ tas_single_node_netcat_image }}"
-    - "{{ tas_single_node_nginx_image }}"
-    - "{{ tas_single_node_tsa_image }}"
+  loop: "{{ images_list }}"
```
bkabrda, line 61: I'm not sure that this change is correct. For example, with your change the Redis image will be the 10th item in the images list and thus will be pulled when `"{{ tas_single_node_tsa_enabled }}"` if I'm looking right.

Both the trillian_db_image and redis_image are already in the list. We just need to update the conditions for them to get pulled when they should be.

bkabrda (approved):
LGTM. Nice job!

PR #16, opened 2024-07-26
chore: Fixes and improvements for the vm-testing setup

bkabrda:
This PR improves/fixes the vm-testing setup:
* Properly overwrite the inventory file regardless where we're calling the provision.sh script from
* Properly set up becoming root in the inventory
* Verify the blob in verbose mode


fghanmi (approved):
lgtm

PR #17, opened 2024-07-26
Replace community.crypto collection usage

bkabrda:
This PR does two things:
* It removes usage of the `community.crypto` collection in favor of `openssl` invocations
* Because we can't/shouldn't install `openssl` on the control node, we install it on the managed node and create all certificates/keys there (they would get copied on there anyway)


fghanmi (approved):
lgtm

PR #18, opened 2024-07-31
Add molecule testing setup using testing farm

bkabrda:
This PR introduces molecule testing setup, using testing-farm to provision VMs. There are a lot of TODOs around idempotency and improvements throughout the molecule setup, but they are ok to use with the simple default scenario.

I also temporarily pinned cosing in the test container to 2.2.4 to workaround TSA certificate chain issues.


fghanmi (approved):
lgtm!

PR #19, opened 2024-08-01
Create leaf certificate for TSA instead of intermediate

bkabrda:
This PR changes the intermediate certificate for TSA (which we don't really need) to be a leaf certificate. Cosign >= 2.3.0 requires that there is a leaf certificate for TSA. I also took the liberty of providing a little more descriptive names for the TSA certificate files.


fghanmi (approved):
lgtm

PR #20, opened 2024-08-01
feat: oidc provider configuration to list

miyunari:
The OIDC config is now provided as a list. 

Jira: https://issues.redhat.com/browse/SECURESIGN-1164

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,8 +75,11 @@ In order to deploy RHTAS on a RHEL 9.2+ VM:
   - hosts: rhtas
     vars:
       base_hostname: TODO # e.g. example.com
-      tas_single_node_oidc_issuers: TODO # your OIDC provider (e.g. keycloak) URL
-      tas_single_node_issuer_url: TODO # your OIDC provider (e.g. keycloak) URL
+      tas_single_node_oidc_issuers:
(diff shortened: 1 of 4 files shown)


bkabrda:
Thanks for the PR, it looks really good. Have you tested usage of multiple issuers? If you did and it worked, I can approve.

osmman:
Can you configure MetaIssuers by OIDC config because Fulcio has two types: OIDCIssuers and MetaIssuers?

miyunari:
@bkabrda yes, I tested it :) 

>Can you configure MetaIssuers by OIDC config because Fulcio has two types: OIDCIssuers and MetaIssuers?
@osmman should we take care of this in an extra PR or do you think it belongs here aswell? :)

osmman:
> > Can you configure MetaIssuers by OIDC config because Fulcio has two types: OIDCIssuers and MetaIssuers?
> 
> @osmman should we take care of this in an extra PR or do you think it belongs here aswell? :)

It can be done in extra PR, it will be good to enable users to set all Fulcio's OIDC options. you can find the definition [here ](https://github.com/securesign/fulcio/blob/main/pkg/config/config.go#L50-L88)

bkabrda (approved):
Approving, thanks! As Tomas noted, please open a followup PR with the additional changes.

PR #21, opened 2024-08-05
Remove unnecessary 'become:true'

fghanmi:


bkabrda:
Have you taken a look at `roles/tas_single_node/tasks/main.yml`? There are some `become: true` there that I think we could remove as well.

fghanmi:
> Have you taken a look at `roles/tas_single_node/tasks/main.yml`? There are some `become: true` there that I think we could remove as well.

yes right, 
This would require the user to always add `ansible_become=true` in the inventory.

bkabrda:
> This would require the user to always add ansible_become=true in the inventory.

Yes, I think we should just mandate this. Most of our tasks will always require that anyway, so I think that shouldn't be a problem.

bkabrda (approved):

PR #22, opened 2024-08-05
feat: extend fulcio configuration for metaissuers

miyunari:
jira: https://issues.redhat.com/browse/SECURESIGN-1283 

follow up PR to: https://github.com/securesign/artifact-signer-ansible/pull/20#issuecomment-2262862402

### DIFF
--- a/roles/tas_single_node/defaults/main.yml
+++ b/roles/tas_single_node/defaults/main.yml
@@ -121,6 +121,23 @@ tas_single_node_oidc_issuers:
     url: https://keycloak-keycloak-system.apps.platform-sts.pcbk.p1.openshiftapps.com/auth/realms/trusted-artifact-signer
     type: email
 
+tas_single_node_meta_issuers:
+  - issuer_pattern: "https://*.oic.prod-aks.azure.com/*"
+    client_id: "sigstore"
+    type: "kubernetes"
+  - issuer_pattern: "https://container.googleapis.com/v1/projects/*/locations/*/clusters/*"
+    client_id: "sigstore"
+    type: "kubernetes"
+  - issuer_pattern: "https://oidc.eks.*.amazonaws.com/id/*"
+    client_id: "sigstore"
+    type: "kubernetes"
+  - issuer_pattern: "https://oidc.prod-aks.azure.com/*"
+    client_id: "sigstore"
+    type: "kubernetes"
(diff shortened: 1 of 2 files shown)


bkabrda (commented):
The change looks good to me, but I'm curious to understand where the default values come from - why this exact list? Is that something that we also set by default in the operator?

bkabrda:
@miyunari thanks for the link! Isn't that just a default config that won't get used by the operator? I thought the operator would create a custom config file where users would have to explicitly set this value and this default config would never get used. I could be totally wrong though :)

osmman:
> @miyunari thanks for the link! Isn't that just a default config that won't get used by the operator? I thought the operator would create a custom config file where users would have to explicitly set this value and this default config would never get used. I could be totally wrong though :)

Hello! You're right that the operator typically generates a custom configuration file where users can explicitly set their Fulcio configuration. The default values you're referring to are likely intended for the public Sigstore instance. To ensure clarity, we want users to explicitly set their Fulcio configuration, so the default values should be an empty list. This approach allows users to define their specific configurations rather than relying on preset defaults. Thank you for pointing this out.

miyunari:
Thanks for explaining and asking @bkabrda and @osmman :) :+1:

bkabrda:
@miyunari Let's make the `defaults/main.yml` contain an explicit empty value for `tas_single_node_meta_issuers`, that's generally a good practice.

miyunari:
> @miyunari Let's make the `defaults/main.yml` contain an explicit empty value for `tas_single_node_meta_issuers`, that's generally a good practice.

@bkabrda yes, that makes sense, I added an empty entry :)

roles/tas_single_node/defaults/main.yml
```diff
@@ -121,6 +121,11 @@ tas_single_node_oidc_issuers:
     url: https://keycloak-keycloak-system.apps.platform-sts.pcbk.p1.openshiftapps.com/auth/realms/trusted-artifact-signer
     type: email
 
+tas_single_node_meta_issuers:
```
bkabrda, line 124: I think we shouldn't add an empty entry, because that might actually be an invalid configuration value. I think we should just use `tas_single_node_meta_issuers: []` - just an empty list to be explicit that there are no meta issuers by default.

miyunari: done :)

bkabrda (approved):

PR #23, opened 2024-08-07
TAS: Ensure the os is supported

miyunari:
Jira: https://issues.redhat.com/browse/SECURESIGN-1246

### DIFF
--- a/roles/tas_single_node/tasks/main.yml
+++ b/roles/tas_single_node/tasks/main.yml
@@ -1,4 +1,7 @@
 ---
+- name: Check OS System Support
+  ansible.builtin.include_tasks: os_support.yml
+
 - name: Install Operating System Components
   ansible.builtin.include_tasks: os.yml
   when: not tas_single_node_skip_os_install
@@ -8,6 +11,3 @@
 
 - name: Execute Podman Actions
   ansible.builtin.include_tasks: podman.yml
-
-- name: Check OS System Support
-  ansible.builtin.include_tasks: os_support.yml
--- a/roles/tas_single_node/tasks/main.yml
+++ b/roles/tas_single_node/tasks/main.yml
@@ -1,5 +1,5 @@
 ---
-- name: Check OS System Support
+- name: Check OS Support
   ansible.builtin.include_tasks: os_support.yml
 
 - name: Install Operating System Components
--- a/roles/tas_single_node/tasks/main.yml
+++ b/roles/tas_single_node/tasks/main.yml
@@ -8,3 +8,6 @@
 
 - name: Execute Podman Actions
   ansible.builtin.include_tasks: podman.yml
+
+- name: Check OS System Support
+  ansible.builtin.include_tasks: os_support.yml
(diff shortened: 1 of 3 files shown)


roles/tas_single_node/tasks/os.yml
```diff
@@ -1,4 +1,22 @@
 ---
+- name: Fail if the OS is not RHEL
+  ansible.builtin.fail:
+    msg: "Unsupported OS. This playbook supports only RHEL."
+  when: ansible_facts['os_family'] != 'RedHat'
+
+- name: Fail if the OS version is lower than 9.2
+  ansible.builtin.fail:
+    msg: "Unsupported OS version. This playbook supports only RHEL 9.2 or higher."
+  when: (ansible_facts['distribution'] == 'RedHat' and
```
bkabrda, line 10: Will this let us run on CentOS as well? I think allowing CentOS is useful as well.

roles/tas_single_node/tasks/os.yml
```diff
@@ -1,4 +1,22 @@
 ---
+- name: Fail if the OS is not RHEL
+  ansible.builtin.fail:
+    msg: "Unsupported OS. This playbook supports only RHEL."
```
bkabrda, line 4: The message should say "role", not "playbook" to be precise. A playbook is something that the user would create and which would apply this role to a (set of) host(s). There are multiple occurences of this in this PR, please remember to fix them all.

bkabrda (changes requested):
Hi, thanks for the PR. It looks ok, but I left some comments inline. One more comment I have is that I think this shouldn't live in the `os.yml` file. This only gets included conditionally from `main.yml`, but we need to run these tasks unconditionally. I would suggest creating a new file that would contain these types of checks and include that from `main.yml`.

miyunari:
@bkabrda thanks for your feedback - I tried to adress all of it :) Please let me know if I missed something or misunderstood maybe :)

roles/tas_single_node/tasks/main.yml
```diff
@@ -8,3 +8,6 @@
 
 - name: Execute Podman Actions
   ansible.builtin.include_tasks: podman.yml
+
+- name: Check OS System Support
```
bkabrda, line 12: We should execute these tasks at the very start, so that we fail fast and users don't observe random failures because things don't work on their system.

bkabrda (changes requested):
Hi, this looks really good, thanks for doing the changes. There is one more thing to fix that I commented on inline, can you please address it?

roles/tas_single_node/tasks/main.yml
miyunari: ah yes, that makes sense! Thanks :+1: I will adress it :)

roles/tas_single_node/tasks/main.yml
```diff
@@ -1,4 +1,7 @@
 ---
+- name: Check OS System Support
```
bkabrda, line 2: Nit: sorry I haven't noticed this before, but let's just use "OS" or "operating system". I will approve the PR, but please fix this before merging. Thanks!

bkabrda (approved):

roles/tas_single_node/tasks/main.yml
miyunari: Yes, good point - I will fix that :)

PR #24, opened 2024-08-13
Update collection and role metadata files and top-level readme

bkabrda:
This PR updates the collection and role metadata files and also the top-level README file with some more up-to-date information. It also removes some old test files that we no longer need.


fghanmi (approved):
LGTM

PR #26, opened 2024-08-19
Don't rely on distribution minor version, it's not always present

bkabrda:
`distribution_minor_version` is not always present, e.g. on CentOS Stream 9. We optimistically assume that when it is missing, we can proceed.


fghanmi (approved):
lgtm

PR #27, opened 2024-08-20
[SECURESIGN-1162] Ensuring no tasks changed on Re-Run of Playbook

tommyd450:
From rerunning the Playbook without making any changes to any of the config the only task that will re-run is the podman login. This task is skipped over if the podman login remains the same.

### DIFF
--- a/roles/tas_single_node/tasks/podman.yml
+++ b/roles/tas_single_node/tasks/podman.yml
@@ -22,6 +22,7 @@
   ansible.builtin.command: podman login registry.redhat.io -u '{{ tas_single_node_registry_username }}' --password {{ tas_single_node_registry_password }}
   register: podman_login_result
   when: logincheck.stdout | string != tas_single_node_registry_username | string
+  changed_when: true
 
 - name: Pull all images
   containers.podman.podman_image:
--- a/roles/tas_single_node/tasks/podman.yml
+++ b/roles/tas_single_node/tasks/podman.yml
@@ -15,8 +15,8 @@
 - name: Podman check login
   ansible.builtin.command: podman login --get-login registry.redhat.io
   register: loginCheck
-  changed_when: False
-  ignore_errors: true
+  failed_when: loginCheck.stdout == "Nil val"
+  changed_when: loginCheck.stdout | string != tas_single_node_registry_username | string
 
 - name: Podman login to registry.redhat.io
   ansible.builtin.command: podman login registry.redhat.io -u '{{ tas_single_node_registry_username }}' --password {{ tas_single_node_registry_password }}
--- a/roles/tas_single_node/tasks/podman.yml
+++ b/roles/tas_single_node/tasks/podman.yml
@@ -12,17 +12,16 @@
     - "{{ tas_single_node_kube_manifest_dir }}"
     - "{{ tas_single_node_kube_configmap_dir }}"
 
-- name: Podman check login
+- name: Check if Podman is already logged in to registry.redhat.io
   ansible.builtin.command: podman login --get-login registry.redhat.io
   register: logincheck
-  failed_when: logincheck.stdout == "Nil val"
+  failed_when: logincheck.stdout == 'Error":" not logged into registry.redhat.io'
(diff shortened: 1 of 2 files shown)


roles/tas_single_node/tasks/podman.yml
```diff
@@ -12,10 +12,17 @@
     - "{{ tas_single_node_kube_manifest_dir }}"
     - "{{ tas_single_node_kube_configmap_dir }}"
 
+- name: Podman check login
+  ansible.builtin.command: podman login --get-login registry.redhat.io
+  register: logincheck
+  failed_when: logincheck.stdout == "Nil val"
```
bkabrda, line 18: In what case would `Nil val` be printed?

roles/tas_single_node/tasks/podman.yml
```diff
@@ -12,10 +12,17 @@
...
+  changed_when: logincheck.stdout | string != tas_single_node_registry_username | string
+
 - name: Podman login to registry.redhat.io
   ansible.builtin.command: podman login registry.redhat.io -u '{{ tas_single_node_registry_username }}' --password {{ tas_single_node_registry_password }}
   register: podman_login_result
-  changed_when: '"Already logged in" not in podman_login_result'
+  when: logincheck.stdout | string != tas_single_node_registry_username | string
+  changed_when: '"Login Succeeded!" in podman_login_result.stdout'
```
bkabrda, line 25: When we run this task, we would always expect it to be changed (because it changes the state of the machine from not being logged in to being logged in), so the `changed_when` can be removed altogether I think.

bkabrda:
@tommyd450 thanks for the PR, it looks good, I left some comments inline to be addressed

Can you please try rebasing on latest main and modifying the `.github/workflows/molecule.yml` file to run `molecule -v test --scenario-name default` instead of the two current molecule commands (converge/verify)? This should make molecule run the `idempotence` stage, and that should verify that your PR is correct, as well as any further PRs in the future.

roles/tas_single_node/tasks/podman.yml
tommyd450: This task is always expected to be changed?
If so i can get rid of the check before hand in that case no?

bkabrda: @tommyd450 this task is always expected to be changed *when it is run*. We just don't want to run it when we're already logged in. Does that make sense?

tommyd450: Right gotcha,

roles/tas_single_node/tasks/podman.yml
```diff
@@ -12,10 +12,17 @@
     - "{{ tas_single_node_kube_manifest_dir }}"
     - "{{ tas_single_node_kube_configmap_dir }}"
 
+- name: Podman check login
```
bkabrda, line 15: Also, please provide a better name for the task - something like "Check whether podman is logged in to the image registry" or similar.

roles/tas_single_node/tasks/podman.yml
```diff
@@ -12,10 +12,16 @@
     - "{{ tas_single_node_kube_manifest_dir }}"
     - "{{ tas_single_node_kube_configmap_dir }}"
 
+- name: Check if Podman is already logged in to registry.redhat.io
+  ansible.builtin.command: podman login --get-login registry.redhat.io
+  register: logincheck
+  failed_when: logincheck.stdout == 'Error":" not logged into registry.redhat.io'
+  changed_when: logincheck.stdout | string != tas_single_node_registry_username | string
+
 - name: Podman login to registry.redhat.io
```
bkabrda, line 21: You probably need to use explicit `changed_when: true` on this task to make ansible-lint happy. In this case is ok, because this task does change host state when it runs (and when it is skipped it won't be changed :))

tommyd450: Ansible lint has been appeased!

tommyd450: awating the molecule run through, which btw is awesome that we have this now.

bkabrda (approved):
LGTM now, thanks!

PR #28, opened 2024-08-20
[SECURESIGN-1281] | Moving Sensitive Variables out of the Defaults main.yml

tommyd450:


roles/tas_single_node/vars/main.yml
```diff
@@ -1,2 +1,70 @@
...
+tas_single_node_scaffolding_utils_image: quay.io/ablock/sigstore-scaffolding-helper:latest
+
+# Timestamp Authority Config
+tas_single_node_signer_type: file
+tas_single_node_remote_tsa_signer_private_key: "{{ tas_single_node_certs_dir }}/{{ tas_single_node_tsa_signer_private_key_filename }}"
+tas_single_node_remote_tsa_leaf_certificate: "{{ tas_single_node_certs_dir }}/{{ tas_single_node_tsa_leaf_certificate_filename }}"
+tas_single_node_remote_tsa_certificate_chain: "{{ tas_single_node_certs_dir }}/{{ tas_single_node_tsa_certificate_chain_filename }}"
+tas_single_node_remote_tsa_private_key: "{{ tas_single_node_certs_dir }}/{{ tas_single_node_tsa_private_key_filename }}"
+
+# RHTAS Images
```
bkabrda, line 46: I would say that we want customers to be able to provide custom image URLs, e.g. if we release a custom build that fixes an issue for a customer etc.

roles/tas_single_node/defaults/main.yml
```diff
@@ -7,6 +7,20 @@ tas_single_node_system_packages:
 
 tas_single_node_podman_network: rhtas
 
+tas_single_node_rekor_templates:
```
bkabrda, line 10: I don't think we want customers to modify these, they serve more as constants rather than configuration values. Let's move them to vars.

roles/tas_single_node/defaults/main.yml
```diff
@@ -43,11 +46,6 @@ tas_single_node_trillian:
     host: trillian-mysql-pod
     port: 3306
 
-tas_single_node_config_dir: /etc/rhtas
-tas_single_node_certs_dir: "{{ tas_single_node_config_dir }}/certs"
-tas_single_node_kube_manifest_dir: "{{ tas_single_node_config_dir }}/manifests"
-tas_single_node_kube_configmap_dir: "{{ tas_single_node_config_dir }}/configs"
-
 tas_single_node_private_key_filename: rhtas.key
```
bkabrda, line 49: I think these will need to change for the following release (we will probably do lists of certificates) and for the first release of the collection we do not promise exposing them. Let's move them all to vars - we'll expose in a future release when we're ready for customers to provide their own certs/keys.

roles/tas_single_node/defaults/main.yml
```diff
@@ -62,89 +60,29 @@ tas_single_node_tsa_certificate_chain_filename: tsa-certificate-chain.pem
...
-tas_single_node_nginx_certs_config: "{{ tas_single_node_kube_configmap_dir }}/nginx-certs.yaml"
-tas_single_node_trillian_secret: "{{ tas_single_node_kube_configmap_dir }}/trillian-secret.yaml"
-tas_single_node_systemd_directory: /etc/systemd/system
 
 tas_single_node_setup_host_dns: true
 tas_single_node_base_hostname: ""
 
-tas_single_node_fulcio_enabled: true
+
 tas_single_node_fulcio_ca_passphrase: rhtas
```
bkabrda, line 70: I would say let's move these to vars as well, same reason as for the certs/keys themselves.

bkabrda (changes requested):
Thanks for doing this, I provided some comments inline to be addressed.

roles/tas_single_node/defaults/main.yml
tommyd450: Noted will make the changes

roles/tas_single_node/defaults/main.yml
tommyd450: Ok will change this.

roles/tas_single_node/vars/main.yml
bkabrda: I meant custom build of the images, not the collection itself. I also think it's possible we would release new versions of images, e.g. 1.1.1 and we don't necessarily need to (immediately) release the Ansible collection, so we just tell customers to upgrade images in the configuration.

tommyd450: ok that makes sense. Will add those now.

bkabrda (changes requested):
Thanks for doing all the changes! Couple more points:
* I forgot that we have the `tas_single_node_system_packages` variable that I don't think we want to expose - if a user changed this, our role would stop working. Let's move it internally.
* For now, I think we also want to move `tas_single_node_cockpit` and set `tas_single_node_cockpit.enabled` to `false`. We're not yet sure whether we will actually do anything about cockpit for the initial release, so let's hide the variables for now.

One question: why is this PR replacing the image urls/shas?

tommyd450:
> One question: why is this PR replacing the image urls/shas?

These are just the sha's from the main branch, my pr was from before they were updated so my sha's were older ones

bkabrda (approved):
LGTM now, thanks! Feel free to merge when the CI passes.

tommyd450:
> LGTM now, thanks! Feel free to merge when the CI passes.

Awesome will do!

PR #30, opened 2024-08-27
Remove an unused template file

bkabrda:
AFAICS this file was committed by mistake a long ago and was never really used.


SequeI (approved):
lgtm

PR #31, opened 2024-08-27
Ensure that all required system dependencies are of 'latest' version

bkabrda:
We already ensure that we're running on RHEL 9.2, but that doesn't necessarily mean that the user has updated all packages. By using `state: latest` we ensure that new enough podman (and other deps) are of the required RHEL 9.2 versions.


lance (approved):
/lgtm

PR #34, opened 2024-08-29
Auto-generate docs for the tas_single_node role

bkabrda:
This PR adds `argument_specs.yml` for the role and adds tooling to regenerate its README.md file.

While at it, I also moved couple more variables from `defaults/main.yml`, because I think we don't want them to be publicly exposed right now.

The linting job now also checks the argument_specs via antsibull and ensures that the role README file has been regenerated if some of its sources changed.


bkabrda:
I know Aron still has some proposals on changing the README template, but I'll merge this PR now to prevent potential conflicts with other PRs. I will open another PR to get the template improved.

PR #36, opened 2024-09-04
[SECURESIGN-1317] | Using tmp ansible module

tommyd450:


roles/tas_single_node/tasks/certificates.yml
```diff
@@ -245,11 +245,18 @@
         (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_tsa_certificate_chain) | list | length) == 0
         and tas_single_node_signer_type == 'file'
 
+
+    - name: Create temporary tmp directory
```
bkabrda, line 249:
```suggestion
    - name: Create temporary directory for manipulating CA certificate chain
```

tommyd450: Done

roles/tas_single_node/tasks/certificates.yml
```diff
@@ -245,13 +245,26 @@
...
       register: result
       changed_when: "'changed' in result.stdout"
       when: >
         (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_tsa_certificate_chain) | list | length) == 0
         and tas_single_node_signer_type == 'file'
+
+    - name: Delete content & directory
+      ansible.builtin.file:
+        state: absent
+        path: tmpdir.path
```
bkabrda, line 270: I think this is incorrect. This should be `{{ tmpdir.path }}` (and you'll find out if that's correct by the idempotence check failing, as this directory will always be present - you'll need to use `changed_when: false` as well here).

tommyd450: Ok will make that change

PR #37, opened 2024-09-04
Reviewed and updated the README content

ritz303:
Reviewed the `README.md` file, and updated its content to align with IBM and Red Hat documentation standards and styles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,212 +1,217 @@
-# RHTAS Ansible Collection
+# Red Hat Trusted Artifact Signer Ansible collection
 
-Automation to deploy the RHTAS ecosystem on RHEL
+The purpose of this Ansible collection is to automate the deployment of the Red Hat Trusted Artifact Signer (RHTAS) service on Red Hat Enterprise Linux (RHEL).
 
-:warning: **The contents of this repository are a Work in Progress.**
+> [!WARNING]
+Currently, the contents of this repository are a work in progress and can change often.
 
 ## Overview
 
-The automation within this repository establishes the components of RHTAS, the downstream redistribution of [Sigstore project](https://sigstore.dev) within a single Red Hat Enterprise Linux (RHEL) machine using a standalone containerized deployment. Containers are spawned using Kubernetes based manifests using
-[podman kube play](https://docs.podman.io/en/latest/markdown/podman-kube-play.1.html).
+The RHTAS service is the downstream redistribution of the [Sigstore](https://sigstore.dev) project.
+The automation contained within this Git repository installs and configures the components of RHTAS to run on a single RHEL server, which uses a standalone containerized deployment.
+A Kubernetes-based manifest creates containers that uses [`podman kube play`](https://docs.podman.io/en/latest/markdown/podman-kube-play.1.html).
 
-The following Sigstore components are deployed as part of this architecture:
+The RHTAS Ansible collection deploys the following RHTAS components:
 
 * [Rekor](https://docs.sigstore.dev/rekor/overview)
-    * [Trillian](https://github.com/google/trillian)
-    * Optionally a MariaDB instance and a Redis instance, although it is possible to use instances managed outside of Ansible
+  * [Trillian](https://github.com/google/trillian)
+  * Optional.
+    A self-managed MariaDB instance, and a Redis instance.
 * [Fulcio](https://docs.sigstore.dev/fulcio/overview)
 * [Certificate Log](https://docs.sigstore.dev/fulcio/certificate-issuing-overview)
 * [Timestamp Authority](https://docs.sigstore.dev/verifying/timestamps/#timestamp-authorities)
-* [TUF](https://theupdateframework.io/) server
+* [TUF](https://theupdateframework.io/)
 
-An [NGINX](https://www.nginx.com) frontend is placed as an entrypoint to the various backend components. Communication is secured via a set of self-signed certificates that are generated at runtime.
+An [NGINX](https://www.nginx.com) front end places an entrypoint to the various backend components.
+A set of self-signed certificates get generated at runtime to establishing secure communications.
 
-Utilize the steps below to understand how to setup and execute the provisioning.
+This automation also deploys and configures a software load balancer as a central point of ingress.
+The ingress host names are as follows, where `<base_hostname>` is your organization's base hostname:
+
+* https://rekor.`<base_hostname>`
+* https://fulcio.`<base_hostname>`
+* https://tsa.`<base_hostname>`
+* https://tuf.`<base_hostname>`
 
 ## Prerequisites
 
-RHEL 9.2+ x86\_64 is supported to run the RHTAS components.
+* RHEL x86\_64 9.2 or greater.
+* Command-line access to the Ansible control node with a user that has `sudo` privileges.
+* Installation and configuration of Ansible on a control node to perform the automation.
+* Installation of the Ansible collections on the control node, unless installing from the Ansible Automation Hub.
+  > [!NOTE]
+  To install the required dependencies automatically, you can run the `ansible-galaxy install` command.
+* An OpenID Connect (OIDC) provider, such as [Keycloak](https://console.redhat.com/ansible/automation-hub/repo/published/redhat/sso/).
+* The ability to resolve the ingress host names, by using the Domain Name System (DNS) or the `/etc/hosts` file.
+* Optional.
+  Installation of the `podman` and [`cosign`](https://github.com/sigstore/cosign) binaries to verify that the RHTAS service is working as expected.
 
-Ansible must be installed and configured on a control node that will be used to perform the automation.
+## Deploying
 
-Perform the following steps to prepare the control node for execution.
+1. Create an `inventory` file with a single node under the `rhtas` group:
+   
+   ```
+   [rhtas]
+   123.123.123.123
+   ```
 
-### Dependencies
+2. Create a simple Ansible Playbook named `play.yml`, and replace `TODO` with your relevant information:
+   
+   ```yaml
+   - hosts: rhtas
+     vars:
+       base_hostname: TODO # e.g. example.com
+       # access credentials for registry.redhat.io (https://access.redhat.com/RegistryAuthentication)
+       tas_single_node_registry_username: TODO
+       tas_single_node_registry_password: TODO
+       tas_single_node_oidc_issuers:
+         - issuer: TODO # your OIDC provider (e.g. keycloak) URL
+           client_id: trusted-artifact-signer
+           url: TODO # your OIDC provider (e.g. keycloak) URL
+           type: email
+     tasks:
+       - name: Include TAS single node role
+         ansible.builtin.include_role:
+           name: redhat.artifact_signer.tas_single_node
+         vars:
+           ansible_become: true
+   ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,212 +1,217 @@
-# RHTAS Ansible Collection
+# Red Hat Trusted Artifact Signer Ansible collection
 
-Automation to deploy the RHTAS ecosystem on RHEL
+The purpose of this Ansible collection is to automate the deployment of the Red Hat Trusted Artifact Signer (RHTAS) service on Red Hat Enterprise Linux (RHEL).
 
-:warning: **The contents of this repository are a Work in Progress.**
+> [!WARNING]
+Currently, the contents of this repository are a work in progress and can change often.
```
bkabrda, line 6: I think we should just go ahead and remove this warning, because we'll release soon and we don't want this to be displayed on Automation Hub. Maybe we should replace it with a note that this is a tech preview?

README.md
```diff
@@ -1,212 +1,217 @@
...
+A Kubernetes-based manifest creates containers that uses [`podman kube play`](https://docs.podman.io/en/latest/markdown/podman-kube-play.1.html).
 
-The following Sigstore components are deployed as part of this architecture:
+The RHTAS Ansible collection deploys the following RHTAS components:
 
 * [Rekor](https://docs.sigstore.dev/rekor/overview)
-    * [Trillian](https://github.com/google/trillian)
-    * Optionally a MariaDB instance and a Redis instance, although it is possible to use instances managed outside of Ansible
+  * [Trillian](https://github.com/google/trillian)
+  * Optional.
```
bkabrda, line 18: Should this be `Optionally` or `Optional: `?

README.md
```diff
@@ -1,212 +1,217 @@
...
+    A self-managed MariaDB instance, and a Redis instance.
 * [Fulcio](https://docs.sigstore.dev/fulcio/overview)
 * [Certificate Log](https://docs.sigstore.dev/fulcio/certificate-issuing-overview)
 * [Timestamp Authority](https://docs.sigstore.dev/verifying/timestamps/#timestamp-authorities)
-* [TUF](https://theupdateframework.io/) server
+* [TUF](https://theupdateframework.io/)
```
bkabrda, line 23: TUF is a framework, I think the ` server` word is actually important here, because "TUF" itself cannot be deployed

README.md
```diff
@@ -1,212 +1,217 @@
...
-Authenticate with the Keycloak instance using the desired credentials.
+## Prerequisites
 
-4. Verify the signed image
+* RHEL x86\_64 9.2 or greater.
+* Command-line access to the Ansible control node with a user that has `sudo` privileges.
+* Installation and configuration of Ansible on a control node to perform the automation.
+* Installation of the Ansible collections on the control node, unless installing from the Ansible Automation Hub.
+  > [!NOTE]
+  To install the required dependencies automatically, you can run the `ansible-galaxy install` command.
```
bkabrda, line 43: This should be `ansible-galaxy install redhat.artifact_signer` (the name `redhat.artifact_signer` could still change, but it's very likely to be this)

README.md
```diff
@@ -1,212 +1,217 @@
...
+       tas_single_node_registry_password: TODO
+       tas_single_node_oidc_issuers:
+         - issuer: TODO # your OIDC provider (e.g. keycloak) URL
+           client_id: trusted-artifact-signer
+           url: TODO # your OIDC provider (e.g. keycloak) URL
+           type: email
+     tasks:
+       - name: Include TAS single node role
+         ansible.builtin.include_role:
+           name: redhat.artifact_signer.tas_single_node
```
bkabrda, line 75: We should leave a comment to use just "tas_single_node" if running from a locally-cloned Git repo.

README.md
```diff
@@ -1,212 +1,217 @@
...
+   $ sudo mv tas-ca.pem /etc/pki/ca-trust/source/anchors/
+   $ sudo update-ca-trust
+   ```
+   > [!TIP]
+   The certificate can be downloaded from the Certificate Viewer by navigating to `https://rekor.<base_domain>` in a web browser.
+   Download the _root_ certificate that issued the Rekor certificate.
+
+## Verifying the deployment by signing a test container
+
+1. Export the following environment variables, replacing `TODO` with your relevant information:
```
bkabrda, line 107: We should also note here that the `/etc/hosts` entries need to be created - same as above.

bkabrda (changes requested):
Hi Aron, thanks for the PR, this looks really awesome. I left a bunch of comments inline to be addressed. I also have one general comment about the code blocks: I see you added the `$` everywhere in the start. I agree that it looks better, but trying to copy the code block with the little button on the right also copies the `$` which makes it impractical for copy-pasting. Could we omit the `$`?

image

README.md
ritz303: Done.

README.md
ritz303: _Optional_ is our standard convention for optional content, but I'll add a colon as you suggested.  For reference, see IBM Style Guide, page 88.

README.md
ritz303: Good point!  Updated accordingly.

README.md
```diff
@@ -1,212 +1,217 @@
...
-* [TUF](https://theupdateframework.io/) server
+* [TUF](https://theupdateframework.io/)
 
-An [NGINX](https://www.nginx.com) frontend is placed as an entrypoint to the various backend components. Communication is secured via a set of self-signed certificates that are generated at runtime.
+An [NGINX](https://www.nginx.com) front end places an entrypoint to the various backend components.
+A set of self-signed certificates get generated at runtime to establishing secure communications.
 
-Utilize the steps below to understand how to setup and execute the provisioning.
+This automation also deploys and configures a software load balancer as a central point of ingress.
+The ingress host names are as follows, where `<base_hostname>` is your organization's base hostname:
```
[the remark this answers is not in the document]
ritz303: Ah, I didn't know that.  Thank you for explaining it!  I've updated the content accordingly, but dropped the word _desired_.

README.md
```diff
@@ -1,212 +1,217 @@
...
-```shell
-cosign sign -y <image>
-```
-
-Authenticate with the Keycloak instance using the desired credentials.
+## Prerequisites
 
-4. Verify the signed image
+* RHEL x86\_64 9.2 or greater.
+* Command-line access to the Ansible control node with a user that has `sudo` privileges.
```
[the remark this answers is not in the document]
ritz303: Hmm... For step 4, we ask the user to use `sudo`, so whatever user account the user logs in with needs to be in the `/etc/sudoers` file.  If the user can run the commands in step 4 without "root" access, then sure, I can drop `sudo` from the prerequisites.  Is the control node where the user needs to run these commands?

README.md
```diff
@@ -1,212 +1,217 @@
...
+* Installation and configuration of Ansible on a control node to perform the automation.
+* Installation of the Ansible collections on the control node, unless installing from the Ansible Automation Hub.
```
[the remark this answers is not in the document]
ritz303: Made this a separate nested bullet point.  I think this should be clearer to the end user.

README.md
ritz303: Good to know, thank you!

README.md
```diff
@@ -1,212 +1,217 @@
...
+  > [!NOTE]
+  To install the required dependencies automatically, you can run the `ansible-galaxy install` command.
+* An OpenID Connect (OIDC) provider, such as [Keycloak](https://console.redhat.com/ansible/automation-hub/repo/published/redhat/sso/).
+* The ability to resolve the ingress host names, by using the Domain Name System (DNS) or the `/etc/hosts` file.
+* Optional.
+  Installation of the `podman` and [`cosign`](https://github.com/sigstore/cosign) binaries to verify that the RHTAS service is working as expected.
```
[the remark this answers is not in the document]
ritz303: True, if the user already has a signed container image to verify, but what if they don't, and they still want to do a simple verification to see if the RHTAS service is working?  Starting with step 3 in the verification section, I've added a simple verification test.  I have the user create  a very simple container (which is empty), and then I ask them to sign and verify that container image.  In this case we do need them to have `podman`.  I'm going to leave this as-is.

README.md
ritz303: What do you mean by "use just 'tas_single_node'"?   Does the user only need to update the `tas_single_node*` fields in this Playbook, if running from a locally cloned Git repo?  How does this differ from users who use this Playbook on Ansible Automation Hub?  Can you provide me examples of what you mean.

README.md
```diff
@@ -1,212 +1,217 @@
...
+
+4. Add the root certificate authority (CA) to your local truststore:
+  
+   ```shell
+   $ sudo openssl x509 -in ~/Downloads/root-cert-from-browser -out tas-ca.pem --outform PEM
+   $ sudo mv tas-ca.pem /etc/pki/ca-trust/source/anchors/
+   $ sudo update-ca-trust
+   ```
+   > [!TIP]
+   The certificate can be downloaded from the Certificate Viewer by navigating to `https://rekor.<base_domain>` in a web browser.
```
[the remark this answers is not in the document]
ritz303: Most organizations have their own DNS servers to do name resolution.  Maybe I'm missing something, but I don't understand why a user would need to update the `/etc/hosts` file, unless they are on an isolated network with no DNS server.  Can we simply say to the user, update your DNS records to point to these servers?

README.md
ritz303: See my comment above.

ritz303:
@bkabrda 
Thank you for taking the time to review my PR!  Ah, I didn't know the copy-and-paste button feature works differently than the same feature on the Customer Portal, so I removed the command prompt (`$`) from the command blocks.  I also made some updates based on your suggestions, and asked for some more clarity on other suggestions.

tommyd450:
@ritz303 This looks good, thanks for throwing this together, outside of whats been addressed I cannot see anything in need of changes.

README.md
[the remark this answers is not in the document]
bkabrda: Oh you're right that we need it in step 4. We need to fix that, but the fix should be different: step 4 should say to do this on all nodes from which the user wants to use the RHTAS client software, like `cosign`/`gitsign`. It's true that this might include the control node, but we should make it clear what this applies to.

README.md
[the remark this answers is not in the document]
bkabrda: Ok, I see what you mean. It's also possible to sign and verify a blob (e.g. a text file) which doesn't require podman at all, but I'm ok if you want to leave this as-is, because signing/verifying container images is likely the most prominent usecase and we want to showcase it.

README.md
bkabrda: So if you're using from a local checkout you'll need to use `name: tas_single_node` here. If you install this collection from AAH via `ansible-galaxy install`, you'll need to use `name: redhat.artifact_signer.tas_single_node`, that's the only difference. Does that make sense?

README.md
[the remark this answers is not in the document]
bkabrda: Yeah, you're right that that's probably the best solution. Modifying `/etc/hosts` is a "quick fix" solution that can be used to quickly test the deployment (as opposed to reaching out to somebody in your company who manages DNS), so maybe we still want to keep it around for the "quick fix" value?

bkabrda:
@ritz303 thanks a lot! I tried to provide as much context as I could to the points you raised, please let me know if my replies make sense.

README.md
ritz303: Yes, that makes sense.  Thank you for clarifying!

PR #40, opened 2024-09-12
Remove old hardcoded OIDC issuer

bkabrda:
This old hardcoded OIDC issuer was useful for testing, but we have it in the Molecule setup and it shouldn't be shipped to customers.


SequeI (approved):
/lgtm

PR #41, opened 2024-09-12
Edits to the README doc

ritz303:
Updates to the root-level README.md file. 

Fixes: Jira [SECURESIGN-49](https://issues.redhat.com/browse/SECURESIGN-49)

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,9 @@ The ingress host names are as follows, where `<base_hostname>` is your deploymen
 ## Prerequisites
 
 * RHEL x86\_64 9.2 or greater.
-* Command-line access to the Ansible control node with a user that has `sudo` privileges.
-* Update DNS records or `/etc/hosts` entries with the ingress host names and IP addresses.
+* All client nodes using `cosign`, `gitsign`, and `ec` need the following:
+  * Command-line access to the node with a user that has `sudo` privileges.
+  * Updated DNS records or `/etc/hosts` entries with the ingress host names and IP addresses.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,181 +1,192 @@
...
-cosign sign -y <image>
-```
-
-Authenticate with the Keycloak instance using the desired credentials.
+## Prerequisites
 
-4. Verify the signed image
+* RHEL x86\_64 9.2 or greater.
+* Command-line access to the Ansible control node with a user that has `sudo` privileges.
+* Update DNS records or `/etc/hosts` entries with the ingress host names and IP addresses.
```
bkabrda, line 42: For the two points above, I still think we need a slightly different wording. The `sudo` access + updating DNS records need to be done on nodes that will use the client software. Note that this wording is quite important because of tools like Ansible Automation Plaftorm, where the whole collection execution must be done without sudo privileges/DNS records etc.

So, can we explicitly say that sudo+DNS is necessary on nodes that will use cosign etc?

bkabrda (changes requested):
Hi Aron, thanks for the PR. I only have one minor note, otherwise it looks great. Once we get that one note sorted out we can merge!

README.md
ritz303: Thank you for that feedback!  What do you think about my latest commit?  Is it clear and accurate?

bkabrda (approved):
LGTM now, thanks! I'm merging.

PR #44, opened 2024-09-18
[SECURESIGN-1370] update fulcio image and configuration

fghanmi:
Modify the Fulcio configuration to align with the updates in the latest Fulcio image: https://github.com/sigstore/fulcio/blob/main/pkg/config/config.go#L53-L119


SequeI (approved):
/lgtm

PR #46, opened 2024-09-23
Changes for Structural Tests

tommyd450:
Renaming of variables to allow for ease of access of strucutural tests to ensure all images are present and accounted for.


bkabrda (approved):
Approving with a note that it would be great to add a comment in the defaults/main.yml file that would tell people who add/change an image variable to ensure that it's consistent with what the structural tests expect.

tommyd450:
> Approving with a note that it would be great to add a comment in the defaults/main.yml file that would tell people who add/change an image variable to ensure that it's consistent with what the structural tests expect.

Awesome note will add that now.

PR #47, opened 2024-09-24
Build collection tarball on tag push

bkabrda:
This PR adds a new workflow that build the collection and stores it as an artifact on tag push.


fghanmi (approved):
LGTM

PR #48, opened 2024-09-24
SECURESIGN-1394 | Add Cli server to ansible

JasonPowr:
What the cli server landing page looks like 
Screenshot from 2024-09-24 12-40-07


roles/tas_single_node/templates/configs/etc-hosts.j2
```diff
@@ -1,3 +1,3 @@
-{% for item in ['fulcio','tuf', 'rekor','rekor-search'] %}
+{% for item in ['fulcio','tuf', 'rekor','rekor-search', 'tsa', 'cli-server'] %}
```
bkabrda, line 1: If we don't end up merging this PR, could you submit a separate PR to add the `tsa` string here? That should be fixed either way.

bkabrda (approved):

PR #49, opened 2024-09-24
dependabot fix

SequeI:
Currently, the dependabot is not working as it does not intake a `file` in it's schema so I removed it, also from my understanding, dependabot with a `pip environment` only detects files ending with a requirements.txt so I changed those to fit that aswell. To see the full error you can go into insights -> dependency graph -> dependabot


bkabrda:
@SequeI can you please explain in the description what problem this PR is solving? Thanks!

SequeI:
Done, sorry forgot that context is important 😅

bkabrda (approved):
LGTM :+1:

PR #50, opened 2024-09-25
Remove hostPort definitions from all pods

bkabrda:
This removes hostPort definitions from all pods - everything that we want to expose should only be exposed through nginx right now. Anything else increases the surface for a potential attacker.

Included is a fix for one of the tasks that was relying on this to fetch the rekor public key.

### DIFF
--- a/roles/tas_single_node/tasks/podman/tuf.yml
+++ b/roles/tas_single_node/tasks/podman/tuf.yml
@@ -16,16 +16,13 @@
   register: remote_tuf_certificates
 
 - name: Retrieve Rekor Public Key
-  ansible.builtin.uri:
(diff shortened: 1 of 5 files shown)


JasonPowr:
The host port is still defined for the nginx server:

https://github.com/securesign/artifact-signer-ansible/blob/slavek.kabrda/no-hostport/roles/tas_single_node/templates/manifests/nginx/nginx.yaml#L28-L34

bkabrda:
@JasonPowr sorry I forgot to mention this in the PR description. The nginx pod is the only one we want to bind to host ports, as we route all the incoming/outgoing traffic through there, so this is expected.

JasonPowr:
LGTM

PR #51, opened 2024-09-25
Fix image build in molecule verify

bkabrda:
For some reason we now need to provide full path to the Containerfile. I'm honestly not sure why.


tommyd450 (approved):

tommyd450:
Sorcery at play here, otherwise it LGTM

PR #59, opened 2024-09-25
deps bump

SequeI:
Bumping dependencies, also adding grouping to dependabot instead of having numerous PR's


testing-requirements.txt
```diff
@@ -1,7 +1,6 @@
-ansible-lint==24.6.1
-antsibull-docs==2.13.1
-galaxy-importer==0.4.25
```
bkabrda, line 3: You accidentally dropped the galaxy-importer dependency which is required now for running the linters job.

bkabrda (approved):

PR #60, opened 2024-09-25
Implement usage of file signer for Rekor

bkabrda:
This replaces the `memory` signer with a `file` signer for Rekor. It also prohibits usage of `memory` signer for now (not useful ATM) and renames some TSA-related variables so it's clear they're meant for TSA.


fghanmi (approved):
LGTM

PR #53, opened 2024-09-25
Drop hard dependency on collections.podman

bkabrda:
We can't have hard dependency on collections.podman as it is not on Ansible Automation Hub right now.

PR #61, opened 2024-09-25
Drop usage of containers.podman collection

bkabrda:
Because the containers.podman collection dependency would make our collection ineligible for certification on Ansible Automation Hub, we need to completely drop it. This PR replaces #53.

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -16,8 +16,7 @@ description: Install and configure RHTAS, a downstream redistribution of the Sig
 license_file: Apache-2.0
 tags: [sigstore, tas, rhtas, security, cosign]
 # NOTE: when updating, also update dependencies in requirements.yml
-dependencies:
-  containers.podman: ">=1.15.0"
+dependencies: {}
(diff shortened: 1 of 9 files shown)


galaxy.yml
```diff
@@ -16,8 +16,7 @@ description: Install and configure RHTAS, a downstream redistribution of the Sig
 license_file: Apache-2.0
 tags: [sigstore, tas, rhtas, security, cosign]
 # NOTE: when updating, also update dependencies in requirements.yml
-dependencies:
-  containers.podman: ">=1.15.0"
+dependencies: {}
 repository: https://github.com/securesign/artifact-signer-ansible/
 documentation: http://TODO.com
```
ritz303, line 21: Are we wanting to add the official product documentation link here?

bkabrda: Oh yeah, I totally forgot about this. Thanks for noticing, I'll open a separate PR, I don't want to mix it in here.

fghanmi (approved):
LGTM

PR #62, opened 2024-09-25
Test using galaxy-importer, fix metadata

bkabrda:
This PR adds a test using galaxy-importer to ensure we are always ready for import to Ansible Automation Hub. It also fixes the rest of the metadata to ensure that the import passes.


fghanmi (approved):
LGTM

PR #64, opened 2024-09-26
RC3 Image Update

tommyd450:


fghanmi (approved):
LGTM

bkabrda (approved):

roles/tas_single_node/tasks/podman/createtree.yml
```diff
@@ -33,6 +33,7 @@
           --admin_server trillian-logserver-pod:8091
       register: createtree_container
       changed_when: true
+      when: tas_single_node_create_tree_enabled | bool
```
bkabrda, line 36: If we want to make this conditional, then the whole block should be conditional, not just this command (because not running it will break subsequent tasks). However, I would probably not add this new variable at all and just do this unconditionally (hence pull the image whenever trillian is enabled)

tommyd450: right on it

bkabrda (approved):

PR #66, opened 2024-09-27
Updating Read me to meet ansible format requirements

tommyd450:
We have been asked by the ansible folks to make some changes to our readme to follow their [template](https://access.redhat.com/articles/7068606).

The only piece of Info I am missing is some proper use cases, I could come up with some but Im sure we have them documented somewhere. 

@agunn303 if you could give this a once over, it would be greatly appreciated, any changes requested ill get to updaitng asap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Before using this collection, you need to install it with the Ansible Galaxy com
 ansible-galaxy collection install redhat.artifact_signer
 ```
 
-You can also include it in a requirements.yml file and install it with `ansible-galaxy collection install -r requirements.yml`, using the format:
+You can also include it in a `requirements.yml` file and install it with `ansible-galaxy collection install -r requirements.yml`, using the format:
 
 
 ```yaml
@@ -267,7 +267,7 @@ Support tickets for RedHat Trusted Artifact Signer can be opened at https://acce
 
 ## Release notes and Roadmap
 
-Release notes can be found [here](https://docs.redhat.com/en/documentation/red_hat_trusted_artifact_signer/1/html/release_notes/index).
+Release notes can be found [here](https://docs.redhat.com/en/documentation/red_hat_trusted_artifact_signer/1.1/html/release_notes/index).
 
 ## Related Information
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Technology Preview features are not supported with Red Hat production service le
 These features provide early access to upcoming product features, enabling customers to test functionality and provide feedback during the development process.
 See the support scope for [Red Hat Technology Preview](https://access.redhat.com/support/offerings/techpreview/) features for more details.
 
-## Overview
+## Description
 
 The RHTAS service is the downstream redistribution of the [Sigstore](https://sigstore.dev) project.
 The automation contained within this Git repository installs and configures the components of RHTAS to run on a single RHEL server, which uses a standalone containerized deployment.
@@ -35,8 +35,10 @@ The ingress host names are as follows, where `<base_hostname>` is your deploymen
 * https://tsa.`<base_hostname>`
 * https://tuf.`<base_hostname>`
 
-## Prerequisites
+## Requirements
 
+* Ansible 2.16.0 or greater
+* Python 3.9.0 or greater
 * RHEL x86\_64 9.2 or greater.
 * All client nodes using `cosign`, `gitsign`, and `ec` need the following:
   * Command-line access to the node with a user that has `sudo` privileges.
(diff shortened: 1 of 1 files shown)


bkabrda:
@tommyd450 I think you missed some of the parts from the template. E.g. in the requirements section it says

```
This section must list the required minimum versions of Ansible and Python, and any Python or external collection dependencies. Include additional information on other prerequisite tasks needed, if applicable.
```

In the installation section there is a large boilerplate that needs to be included.

The usage should be moved to the "use cases" section I believe.

Can you please go through the template again and ensure all of the requirements noted there are satisfied in our readme? Thanks!

README.md
```diff
@@ -35,8 +35,10 @@ The ingress host names are as follows, where `<base_hostname>` is your deploymen
 * https://tsa.`<base_hostname>`
 * https://tuf.`<base_hostname>`
 
-## Prerequisites
+## Requirements
 
+* Ansible 2.16.0 or greater
+* Python 3.12.4 or greater
```
bkabrda, line 41: RHEL 9 has 3.9 and I think there's no RHEL version with a newer Python version. I would suggest requiring >= 3.9 (TBH we don't ship any Python code so we could go even below that... but I guess 3.9 is a reasonable requirement).

README.md
```diff
@@ -50,7 +52,39 @@ The ingress host names are as follows, where `<base_hostname>` is your deploymen
...
+## Installation
+
+
+Before using this collection, you need to install it with the Ansible Galaxy command-line tool:
+
+```
+ansible-galaxy collection install redhat.artifact_signer
+```
+
+You can also include it in a requirements.yml file and install it with ansible-galaxy collection install -r requirements.yml, using the format:
```
bkabrda, line 64: Could we use code formatting for the filenames/commands, rather than regular text?

README.md
```diff
@@ -50,7 +52,39 @@ The ingress host names are as follows, where `<base_hostname>` is your deploymen
...
+
+```
+ansible-galaxy collection install redhat.artifact_signer:==1.1.0
+```
+
+See [using Ansible collections](https://docs.ansible.com/ansible/devel/user_guide/collections_using.html) for more details.
+
+
 
 1. Create an `inventory` file with a single node under the `rhtas` group:
```
bkabrda, line 89: I think from this line on, we want to use the `Use Cases` section, Also, please move the `Contributing` section below `Testing`.

README.md
```diff
@@ -50,69 +52,39 @@ The ingress host names are as follows, where `<base_hostname>` is your deploymen
...
+```
+ansible-galaxy collection install redhat.artifact_signer
+```
 
-   - If running from a locally-cloned Git repository, then run the following command:
-   
-     ```shell
-     export ANSIBLE_ROLES_PATH="roles/" ; ansible-playbook -i inventory play.yml
-     ```
+You can also include it in a requirements.yml file and install it with `ansible-galaxy collection install -r requirements.yml`, using the format:
```
ritz303, line 64: Put back ticks around "requirements.yml".

requirements.yml --> `requirements.yml`

README.md
```diff
@@ -50,69 +52,39 @@ The ingress host names are as follows, where `<base_hostname>` is your deploymen
...
+
+
+```yaml
+collections:
+  - name: redhat.artifact_signer
+```
+
+Note that if you install any collections from Ansible Galaxy, they will not be upgraded automatically when you upgrade the Ansible package.
```
ritz303, line 72: We don't want to use passive voice in technical documentation.  Please rewrite this sentence in active voice.  For example:

Installing any collections from Ansible Galaxy does not upgrade the Ansible packages automatically.

README.md
```diff
@@ -50,69 +52,39 @@ The ingress host names are as follows, where `<base_hostname>` is your deploymen
...
+To upgrade the collection to the latest available version, run the following command:
+
+```
+ansible-galaxy collection install redhat.artifact_signer --upgrade
+```
+
+You can also install a specific version of the collection, for example, if you need to downgrade when something is broken in the latest version (please report an issue in this repository). Use the following syntax to install version 1.1.0:
```
ritz303, line 79: We should not use parenthetical statements in technical documentation, nor be so formal.  I would take the parenthetical statement and make it, its own sentence.  For example:

... in the latest version.  Report any problems by filing an issue in this GitHub project.  Use the following ...

README.md
```diff
@@ -223,7 +259,27 @@ ansible-test sanity
...
 
+## Contributing
+
+## Support
+
+Support tickets for RedHat Trusted Artifact Signer can be opened at https://access.redhat.com/support/cases/#/case/new?product=Red%20Hat%20Trusted%20Artifact%20Signer.
+
+## Release notes and Roadmap
+
+Release notes can be found [here](https://docs.redhat.com/en/documentation/red_hat_trusted_artifact_signer/1/html/release_notes/index).
```
ritz303, line 270: This is the wrong link to the official release notes for RHTAS 1.1.  The new link will be --> https://docs.redhat.com/en/documentation/red_hat_trusted_artifact_signer/1.1/html-single/release_notes/index 

This link currently gives you a 404 page error because the RHTAS 1.1 Release Notes have not been published yet, but this link will work once the release notes get published on release day.

README.md
bkabrda: Hi Aron, unfortunately this is a part of the boilerplate that the Ansible team wants to have as-is in all readmes. We can bring this suggestion with them, but I think for now we have to keep this.

README.md
bkabrda: (Same as above, this is what the Ansible team requires as-is right now)

README.md
tommyd450: Awesome thanks

bkabrda (approved):

PR #69, opened 2024-10-02
Properly template rekor URL in rekor-search configuration

bkabrda:
This was missed in the original review #10. The URL needs to be templated so that rekor-search UI properly points to the deployed rekor.


SequeI (approved):
/lgtm

PR #71, opened 2024-10-02
Ensure test fails if TSA cert is not distributed through TUF

bkabrda:
By passing `--use-signed-timestamps` instead of explicitly providing the TSA cert chain, we verify that cosign got the TSA cert chain through TUF.


SequeI (approved):
/lgtm

PR #79, opened 2024-10-22
user provisioned components scenario

SequeI:


.github/workflows/molecule.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
             # ensure we wipe out any data potentially still on the runner from previous run
             molecule reset
             # NOTE: for now we don't run "molecule test", because we don't have proper idempotence yet
-            molecule -v test --scenario-name default
+            molecule -v test --scenario-name user_provided
```
bkabrda, line 36: (This might be just a leftover from testing this scenario, but in general we don't want to replace the default scenario but run both of them; we could even do that in separate jobs by using a [matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#using-a-matrix-strategy))

molecule/user_provided/converge.yml
```diff
@@ -0,0 +1,20 @@
+---
```
bkabrda, line 1: Are some of these create/converge/... files literal C&P? If so, we should be able to use relative symlinks instead. We should only create the new files assuming something needs to be changed.

I think the structure of the default scenario should allow us to only provide a different vars.yml files.

molecule/user_provided/prepare.yml
```diff
@@ -0,0 +1,81 @@
+---
+- name: Prepare
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  become: yes
+  tasks:
+    - name: stopping service
```
bkabrda, line 8: Couple points here that are not specific to just this task:
* Let's capitalize the task names
* Let's make task names more descriptive, it helps to understand the failure (e.g. "Stop the mariadb service" would be a better name)

molecule/user_provided/prepare.yml
```diff
@@ -0,0 +1,81 @@
...
+      ansible.builtin.service:
+        name: mariadb
+        state: stopped
+
+    - name: uninstalling
+      ansible.builtin.yum:
+        name:
+          - mariadb
+          - mariadb-server
+          - redis
```
bkabrda, line 18: I guess I understand that you want to make sure that there is no preexisting initialized mariadb instance that would interfere with the test; however, I'm not sure why you would uninstall the packages, because you can just rely on the install tasks with "state: latest" that you use later to upgrade to latest version. I thing uninstalling is redundant here.

molecule/user_provided/prepare.yml
```diff
@@ -0,0 +1,81 @@
...
+        state: absent
+
+    - name: deletings mysql
+      file:
+        path: /var/lib/mysql
+        state: absent
+
+    - name: Install Redis
```
bkabrda, line 26: I would suggest installing all packages in one task; it doesn't sacrifice readability and runs fewer dnf transactions, which makes the execution much faster.

molecule/user_provided/prepare.yml
```diff
@@ -0,0 +1,81 @@
...
+        name: mariadb
+        enabled: yes
+        state: started
+
+    - name: Create Trillian Database
+      shell: echo "CREATE DATABASE trillian;" | mysql
+
+    - name: Download schema
+      shell: |
+        curl -sSL  https://raw.githubusercontent.com/securesign/trillian/refs/heads/main/storage/mysql/schema/storage.sql > schema.sql
```
bkabrda, line 64: I would suggest always using `curl --retry` to ensure that network/infrastructure flakiness doesn't make this fail. However, since this is Ansible, you should use https://docs.ansible.com/ansible/latest/collections/ansible/builtin/get_url_module.html plus retry/delay https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html#retrying-a-task-until-a-condition-is-met

molecule/user_provided/vars/vars.yml
```diff
@@ -0,0 +1,25 @@
...
+tas_single_node_registry_username: "{{ lookup('env', 'TAS_SINGLE_NODE_REGISTRY_USERNAME') }}"
+tas_single_node_registry_password: "{{ lookup('env', 'TAS_SINGLE_NODE_REGISTRY_PASSWORD') }}"
+tas_single_node_trillian:
+  database_deploy: false
+  mysql:
+    user: mysql
+    root_password: rootpassword
+    password: password
+    database: trillian
+    host: localhost
```
bkabrda, line 18: Does localhost work here? I don't understand how it would, because that would make the trillian container try to reach itself rather than the host that it's running on. I think this should be the molecule host IP, right (same for Redis)?

bkabrda (changes requested):
Hi Aleksy. thanks for submitting the PR. I think it's a great start, but I posted some comments inline to get addressed before this can be merged. Can you please address these?

molecule/user_provided/prepare.yml
```diff
@@ -0,0 +1,62 @@
+---
+- name: Prepare
+  hosts: molecule
+  gather_facts: true
+  become: yes
+  vars_files:
+    - vars/vars.yml
+  tasks:
+    - name: Installing packages
```
bkabrda, line 9:
```suggestion
    - name: Install Redis and MariaDB
```

molecule/user_provided/prepare.yml
```diff
@@ -0,0 +1,62 @@
...
+      ansible.builtin.service:
+        name: redis
+        enabled: yes
+        state: started
+
+    - name: Set Redis password
+      shell: |
+        echo "CONFIG SET requirepass "password"" | redis-cli
+
+    - name: Uncommenting bind-address
```
bkabrda, line 27:
```suggestion
    - name: Allow external connections for Redis
```

bkabrda (approved):
This looks great, thanks for doing the changes. I'm approving with some minor suggestions - I think they improve how a user would read and understand (and potentially troubleshoot) the tasks (yes, this is just testing setup, but let's try to get these things 100 % as well :)).

PR #86, opened 2024-10-31
Use tuffer to pre-initialize the TUF repository and serve it via an httpd container

bkabrda:
This PR makes the collection use the "tuffer" image to pre-create the TUF repository first and then use a simple httpd container to serve the repository.

Contained are workarounds for features missing in podman 4.4 (our baseline) - inability to create k8s "Job" (replaced by a simple Deployment with "restartPolicy: Never" and inability to propagate exit code of a Deployment to systemd (checking for container exit code and failing with an error message if the exit code != 0).


SequeI (approved):
LGTM, please add port and pod/service name as a variable for new TUF addition! Thanks

SequeI:
LGTM after lint passes, thanks for taking a recent manifest with all the probes included !

tommyd450:
Installing playbook and poking around now! I wish to look at changes within the machine im running it on.

tommyd450:
Following up on this, started an install and oddly the httpd server isnt there, podman logs report its creation yet there is no pod/container? Running the install on an Openstack Machine.

bkabrda:
@tommyd450 I provided an explicit unique hostPort for the TUF server, so that should fix this PR.

Context: we discovered in an offline investigation that some old podman versions, e.g. 4.4.1, automatically expose the container port as hostPort, so we must either have unique container ports (impossible in this case, the httpd container we're using hardcodes 8080) or we have to explicitly provide a unique hostPort to do which is what I did.

tommyd450:
@bkabrda Having looked over it earlier it all looks good to me!

PR #88, opened 2024-11-01
Removing leftovers from k8s + refactoring manifests

SequeI:


roles/tas_single_node/vars/main.yml
```diff
@@ -90,3 +90,48 @@ tas_single_node_remote_tsa_signer_private_key: "{{ tas_single_node_certs_dir }}/
 tas_single_node_remote_tsa_leaf_certificate: "{{ tas_single_node_certs_dir }}/{{ tas_single_node_tsa_leaf_certificate_filename }}"
 tas_single_node_remote_tsa_certificate_chain: "{{ tas_single_node_certs_dir }}/{{ tas_single_node_tsa_certificate_chain_filename }}"
 tas_single_node_remote_tsa_private_key: "{{ tas_single_node_certs_dir }}/{{ tas_single_node_tsa_private_key_filename }}"
+
+# Pod/Service names + Ports
+tas_single_node_cli_server_pod: cli-server
+tas_single_node_cli_server_port: 8080
+
+tas_single_node_ctlog_pod: ctlog
+tas_single_node_ctlog_port1: 6962
```
bkabrda, line 99: I think instead using "port1", "port2" etc, we should name these ports - e.g. is this a gRPC port, http port, something else? I assume they all have separate different usage and naming them based on that is more useful.

bkabrda (changes requested):
This must have been extremely tedious, so thank you very much for taking the time to do all of it! I just have one inline comment to address related to naming of several variables. Once that is addressed I will approve.

PR #89, opened 2024-11-01
feat: Add support to pull images from container mirrors

osmman:


.github/workflows/molecule.yaml
```diff
@@ -19,9 +19,6 @@ jobs:
             pip install -r requirements-testing.txt
             ansible-galaxy install -r requirements.yml
             ansible-galaxy install -r molecule/requirements.yml
-        - name: Image mirroring
-          run: |
-            sed -f vm-testing/dev-images.sed  -i roles/tas_single_node/defaults/main.yml
```
lance, line 24: You can probably remove this file, right? https://github.com/securesign/artifact-signer-ansible/blob/main/vm-testing/dev-images.sed

osmman: I will keep that script there because it is still referenced in doc for manual testing.

I modified molecule doc to remove that step.

molecule/default/vars/podman.yml
```diff
@@ -0,0 +1,27 @@
+tas_single_node_podman:
+  policy: "{{ lookup('env', 'MOLECULE_SCENARIO_DIRECTORY') }}/../../vm-testing/containers/policy.json"
```
bkabrda, line 2: In general, a better approach is having the variable contain the contents of the file as opposed to path to the file. This works better with AAP execution environments where users would more likely specify the value rather than handle creating the file.

If users want to source the value from a file, they can use https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_lookup.html

bkabrda (changes requested):
This is really cool, I didn't know about this feature. Thanks for contributing!

Could you make this feature accessible/documented for users? It would involve modifying defaults/main.yml to add empty values for these and meta/argument_specs.yml to add description of the new configuration values. (If you don't have time to do this right now, give me a shout, we can merge without it and I'll tackle it in a separate PR).

This might also need rebase and modification for the new molecule scenario in https://github.com/securesign/artifact-signer-ansible/pull/79 so that this new feature gets used there as well.

molecule/default/vars/podman.yml
osmman: I changed tas_single_node_podman.policy to simply contains policy.json content and extended meta/argument_specs.yml schema

bkabrda (approved):
Awesome, thanks! LGTM.

PR #94, opened 2024-11-05
Updating manifests 

SequeI:
* Updating manifests with missing checks so it is exactly the same as the operator
* Didn't carry over the rekor-sharding volume as it's already it's own ticket
* Added new labels that semantically make sense, such as `part-of`, and decided on a general label scheme for our ansible deployment which is `podman.io`

Currently podman kube play only supports livenessProbes, but I also added readinessProbes if support is added in the future, plus doesn't harm the actual pod deployment in anyway

### DIFF
--- a/roles/tas_single_node/templates/manifests/cli-server/cli-server.j2
+++ b/roles/tas_single_node/templates/manifests/cli-server/cli-server.j2
@@ -4,20 +4,20 @@ metadata:
   name: "{{ tas_single_node_cli_server_pod }}"
   namespace: cli-server
   labels:
-    app.kubernetes.io/component: client-server
+    app.podman.io/component: client-server
     app.name: client-server
 spec:
   replicas: 1
   selector:
     matchLabels:
-      app.kubernetes.io/component: client-server
-      app.kubernetes.io/part-of: trusted-artifact-signer
+      app.podman.io/component: client-server
+      app.podman.io/part-of: trusted-artifact-signer
       app.name: client-server
   template:
     metadata:
       labels:
-        app.kubernetes.io/component: client-server
-        app.kubernetes.io/part-of: trusted-artifact-signer
+        app.podman.io/component: client-server
+        app.podman.io/part-of: trusted-artifact-signer
         app.name: client-server
     spec:
       containers:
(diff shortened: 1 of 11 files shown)


tommyd450:
We havent had the readiness probes prior, what specifically are they doing? This is for my own knowledge. Other wise LGTM

tommyd450 (approved):

bkabrda:
Can you please provide more info on label naming? In https://github.com/securesign/artifact-signer-ansible/pull/88 you changed them and now you seem to be partially changing back. Is there a particular difference? What effect does this have?

SequeI:
> We havent had the readiness probes prior, what specifically are they doing? This is for my own knowledge. Other wise LGTM

A GET request is sent to some endpoint, and depending on what code is returned (200/400) we will know if the pod is up or down, in a cursory check. (The pod can be up, but the service can be down in some circumstances) but it serves as a decent check before wasting any more resources on a pod that is not even up or ready to be interacted with.

SequeI:
> Can you please provide more info on label naming? In #88 you changed them and now you seem to be partially changing back. Is there a particular difference? What effect does this have?

@bkabrda In the last PR I had removed labels that we don't even use anymore (Helm/scaffold) and changed some of the names to fit our purposes better for filtering if a user has a wider ecosystem running than only TAS. 

This time as I was making sure our manifests are up to date I noticed Tomas had recently changed the labels to  be more semantic (Product name label `trusted-artifact-signer` etc) which I included within this PR. Some of the names were just not very descriptive too (Such as `rekor-search-ui` being just `search`).

 It's mainly for easy filtering which could be handy if in the future we go down the route of HA and have many more pods/services running I think. lmk your thoughts though as this is still a very fresh area for me.

bkabrda:
@SequeI thanks for the context. I mostly meant the changes from `app.kubernetes.io/component` to `app.component` and back, is that consequential for usage or is it just to get consistent with the operator? (I want to do the right thing in https://github.com/securesign/artifact-signer-ansible/pull/86 before I merge it, so I'm trying to understand what exactly to use)

bkabrda:
@SequeI thanks, that makes sense. One more question: some containers got `securityContext` to drop privileges, but some didn't. I don't see a clear distinction/reason for this. For the containers that didn't get this argument, could you specify why in a code comment?

bkabrda:
@SequeI any movement on this?

SequeI:
@bkabrda I tried asking around why originally it is present in some manifests and not in others, but didn't really get any concrete answers. I would imagine it was just dev preference during creation and abiding to principle of least privilege. Adding it would be purely to stay 1:1 with what operator containers do

bkabrda (approved):

PR #96, opened 2024-11-06
Configuring non root sudoer for molecule tests

SequeI:
Adding Redis config which I forgot in initial scenario PR for user_provided
Adding non-root sudoer configuration for our scenario tests to emulate a real user environment
Changed shell commands to ansible collections, such as lineinfile

(Waiting for tf access so I can personally snoop around our molecule scenarios on a tf vm before merge)

### DIFF
--- a/molecule/default/prepare.yml
+++ b/molecule/default/prepare.yml
@@ -2,9 +2,18 @@
 - name: Prepare
   hosts: molecule
   gather_facts: false
+  vars:
+    users:
+      - user: testingUser
+        password: testingPassword0x
   vars_files:
     - vars/vars.yml
 
   tasks:
+
     - name: Configure Dex OIDC instance
       ansible.builtin.include_tasks: ../dex-config.yaml
+
+    - name: Setup a non-root sudoer to replicate a user environment
+      ansible.builtin.include_tasks: ../testing_user_setup.yaml
+      with_items: "{{ users }}"
--- a/molecule/default/prepare.yml
+++ b/molecule/default/prepare.yml
@@ -5,7 +5,7 @@
   vars:
     users:
       - user: testingUser
-        password: testingPassword0x
+        password: password123
   vars_files:
     - vars/vars.yml
 
--- a/molecule/testing_user_setup.yaml
+++ b/molecule/testing_user_setup.yaml
@@ -1,13 +1,17 @@
 ---
+- name: Make sure Passlib is present
+  ansible.builtin.pip:
+    name: passlib
+    state: present
+  delegate_to: localhost
+
 - name: Create a non-root sudoer user
   ansible.builtin.user:
     name: "{{ item.user }}"
     shell: /bin/bash
     group: wheel
+    password: "{{ item.password | password_hash }}"
     create_home: yes
-
-- name: Set password for testingUser
-  ansible.builtin.shell: echo {{ item.password }} | passwd --stdin {{ item.user }}
 
 - name: Configure Sudoers for the user
   ansible.builtin.lineinfile:
--- a/molecule/testing_user_setup.yaml
+++ b/molecule/testing_user_setup.yaml
@@ -1,0 +1,51 @@
+---
+- name: Create a non-root sudoer user
+  ansible.builtin.user:
+    name: "{{ item.user }}"
+    shell: /bin/bash
+    group: wheel
+    create_home: yes
+
+- name: Set password for testingUser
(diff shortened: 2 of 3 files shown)


molecule/default/vars/vars.yml
```diff
@@ -8,3 +8,4 @@ tas_single_node_cockpit:
   enabled: false
 tas_single_node_registry_username: "{{ lookup('env', 'TAS_SINGLE_NODE_REGISTRY_USERNAME') }}"
 tas_single_node_registry_password: "{{ lookup('env', 'TAS_SINGLE_NODE_REGISTRY_PASSWORD') }}"
+tas_single_node_user: testingUser
```
bkabrda, line 11: Can we rename this variable? It makes it seem like it's part of the collection configuration and could confuse users who are looking for examples. I would suggest dropping the role prefix and also adding a code comment that this is used by molecule when setting up the VM.

molecule/default/prepare.yml
```diff
@@ -0,0 +1,58 @@
...
+    - vars/vars.yml
+
+  tasks:
+    - name: Create a non-root sudoer user
+      ansible.builtin.user:
+        name: "{{ tas_single_node_user }}"
+        shell: /bin/bash
+        create_home: yes
+
+    - name: Grant passwordless sudo to new user
```
bkabrda, line 17: Would it be possible to not use passwordless? That's closer to a real world scenario.

molecule/default/prepare.yml
```diff
@@ -0,0 +1,58 @@
...
+        path: /etc/ssh/sshd_config
+        regexp: '^(#*)?PermitRootLogin'
+        line: PermitRootLogin no
+      notify: Restart SSH
+
+    - name: Modify Molecule inventory
+      ansible.builtin.lineinfile:
+        path: "{{ molecule_ephemeral_directory }}/inventory/molecule_inventory.yml"
+        regexp: 'ansible_user:'
+        line: "ansible_user: {{ tas_single_node_user }}}"
```
bkabrda, line 48: The three `}` chars at the end look like a typo (?)

molecule/user_provided/prepare.yml
```diff
@@ -60,3 +70,52 @@
         GRANT ALL ON trillian.* to 'mysql'@'%' IDENTIFIED BY 'password' WITH GRANT OPTION;
         FLUSH PRIVILEGES;
         EOF
+
+    - name: Create a non-root sudoer user
```
bkabrda, line 74: Why not create a file that does the user setup and then include it in the prepare files for both of the scenarios?

bkabrda (changes requested):
Hi, thanks a lot for the PR. It looks good overall, but I left some comments inline to address. Can you PTAL?

molecule/default/vars/vars.yml
SequeI: Renamed to testing_user + added comment

molecule/default/prepare.yml
SequeI: This would be the ending of a json object , so the additional } is to actually close it off

molecule/user_provided/prepare.yml
SequeI: Will do, never good to duplicate code

bkabrda:
It seems that something is wrong with the way password is provided/used. From the failed task:

```
  fatal: [3.142.246.43]: FAILED! => {"msg": "Incorrect sudo password"}
```

SequeI:
OIDC instance down so CI won't pass + createtree container fails to spin up on centOS but is fine on RHEL. Will test again once identity instance is back to remove an outside factor in failure but all works as it should on a RHEL machine.

molecule/testing_user_setup.yaml
```diff
@@ -0,0 +1,51 @@
+---
+- name: Create a non-root sudoer user
+  ansible.builtin.user:
+    name: "{{ item.user }}"
+    shell: /bin/bash
+    group: wheel
+    create_home: yes
+
+- name: Set password for testingUser
+  ansible.builtin.shell: echo {{ item.password }} | passwd --stdin {{ item.user }}
```
SequeI, line 10: Manually setting the password, using user.password previously resulted in constant password errors for some reason

SequeI:
I constantly ran into the createtree issue before today, but after rebasing with the new Dex OIDC config + changing a small discrepancy it seems to work fine? Still not sure what the underlying issue was

bkabrda (approved):
Nice job, LGTM. Thanks!

PR #100, opened 2024-11-12
SECURESIGN-1556 | Two log entries being made in Rekor

JasonPowr:
Fix for SECURESIGN-1556, two entries being pushed to the rekor log for a single signature, caused by ctlog and rekor configured to use the same tree in trillian


fghanmi (approved):
LGTM!

PR #101, opened 2024-11-13
SECURESIGN-1476 | Add the Redis backfill job to Ansible collection

JasonPowr:
Pr to add the backfill redis, setting as draft for now as it should only be merged after https://github.com/securesign/artifact-signer-ansible/pull/100

Upstream pr: https://github.com/sigstore/rekor/pull/2280

### DIFF
--- a/roles/tas_single_node/README.md
+++ b/roles/tas_single_node/README.md
@@ -19,7 +19,7 @@ Deploy the [RHTAS](https://docs.redhat.com/en/documentation/red_hat_trusted_arti
 |---|---|---|---|
 | tas_single_node_podman_network | Name of the Podman network for containers to use. | str |  `rhtas`  |
 | tas_single_node_rekor_redis | Details on the Redis connection for Rekor. You can set this to a custom Redis instance. | dict of 'tas_single_node_rekor_redis' options |  `{'database_deploy': True, 'redis': {'host': 'rekor-redis-pod', 'port': 6379, 'password': 'password'}}`  |
-| tas_single_node_backfill_redis_enabled | Enable or disable the backfill redis job | bool |  `True`  |
+| tas_single_node_backfill_redis | Configuration options for the backfill redis job. | dict of 'tas_single_node_backfill_redis' options |  `{'enabled': True, 'schedule': '*-*-* 00:00:00'}`  |
 | tas_single_node_trillian | Details on the database connection for Trillian. You can set this to a custom MySQL or MariaDB instance. | dict of 'tas_single_node_trillian' options |  `{'database_deploy': True, 'mysql': {'user': 'mysql', 'root_password': 'rootpassword', 'password': 'password', 'database': 'trillian', 'host': 'trillian-mysql-pod', 'port': 3306}}`  |
 | tas_single_node_rekor_public_key_retries | The number of attempts to retrieve the Rekor public key when constructing the trust root. | int |  `5`  |
 | tas_single_node_rekor_public_key_delay | The number of seconds to wait before retrying the retrieval of the Rekor public key when constructing the trust root. | int |  `10`  |
@@ -61,6 +61,13 @@ Deploy the [RHTAS](https://docs.redhat.com/en/documentation/red_hat_trusted_arti
 | host | The Redis host. | str | no |  |
 | port | The Redis host port number. | int | no |  |
 | password | The Redis password. | str | no |  |
+
+#### Options for main > tas_single_node_backfill_redis
+
+|Option|Description|Type|Required|Default|
+|---|---|---|---|---|
+| enabled | Enable or disable the backfill redis job. | bool | no |  |
+| schedule | Schedule the backfill redis job should follow. | str | no |  |
 
 #### Options for main > tas_single_node_trillian
 
--- a/roles/tas_single_node/README.md
+++ b/roles/tas_single_node/README.md
@@ -19,6 +19,7 @@ Deploy the [RHTAS](https://docs.redhat.com/en/documentation/red_hat_trusted_arti
 |---|---|---|---|
 | tas_single_node_podman_network | Name of the Podman network for containers to use. | str |  `rhtas`  |
 | tas_single_node_rekor_redis | Details on the Redis connection for Rekor. You can set this to a custom Redis instance. | dict of 'tas_single_node_rekor_redis' options |  `{'database_deploy': True, 'redis': {'host': 'rekor-redis-pod', 'port': 6379, 'password': 'password'}}`  |
+| tas_single_node_backfill_redis_enabled | Enable or disable the backfill redis job | bool |  `True`  |
 | tas_single_node_trillian | Details on the database connection for Trillian. You can set this to a custom MySQL or MariaDB instance. | dict of 'tas_single_node_trillian' options |  `{'database_deploy': True, 'mysql': {'user': 'mysql', 'root_password': 'rootpassword', 'password': 'password', 'database': 'trillian', 'host': 'trillian-mysql-pod', 'port': 3306}}`  |
 | tas_single_node_rekor_public_key_retries | The number of attempts to retrieve the Rekor public key when constructing the trust root. | int |  `5`  |
 | tas_single_node_rekor_public_key_delay | The number of seconds to wait before retrying the retrieval of the Rekor public key when constructing the trust root. | int |  `10`  |
--- a/roles/tas_single_node/templates/manifests/rekor/backfill_redis.j2
+++ b/roles/tas_single_node/templates/manifests/rekor/backfill_redis.j2
@@ -29,16 +29,40 @@ spec:
             - /bin/sh
             - -c
           args:
-            - >
-              endIndex=$(curl -sS http://{{ tas_single_node_rekor_server_pod }}-pod:{{ tas_single_node_rekor_server_port_http }}/api/v1/log | sed -E 's/.*"treeSize":([0-9]+).*/\1/');
-              endIndex=$((endIndex-1));
+            - |
+              endIndex=$(curl -sS http://{{ tas_single_node_rekor_server_pod }}-pod:{{ tas_single_node_rekor_server_port_http }}/api/v1/log | sed -E 's/.*"treeSize":([0-9]+).*/\1/')
+              endIndex=$((endIndex-1))
+
               if [ "${endIndex}" -lt 0 ]; then
-                echo "info: no rekor entries found";
-                exit 0;
-              fi;
-              backfill-redis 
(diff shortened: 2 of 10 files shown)


molecule/user_provided/vars/vars.yml
```diff
@@ -23,3 +23,11 @@ tas_single_node_rekor_redis:
     host: "{{ ansible_default_ipv4.address }}"
     port: 6379
     password: password
+  backfill_redis:
```
SequeI, line 26: Not sure there is a reason to leave these vars here unless you change something, since they are exactly the same as in `default`. Unless you aim to change these into something that will override defaults.

roles/tas_single_node/templates/manifests/rekor/backfill_redis.sh.j2
```diff
@@ -0,0 +1,38 @@
...
+fi
+
+if [ "$startIndex" -gt "$endIndex" ]; then
+  startIndex=0
+fi
+
+if [ "$startIndex" -le "$endIndex" ]; then
+  podman run --rm --name backfill_redis \
+    -v "{{ tas_single_node_certs_dir }}/ingress-rekor.pem:/usr/local/share/ca-certificates/ingress-rekor.pem:ro" \
+    -e SSL_CERT_FILE=/usr/local/share/ca-certificates/ingress-rekor.pem \
```
SequeI, line 25: Is this the correct filepath? iirc the path for ingress certs is `/etc/rhtas/certs/ingress-{{component}}.pem`

SequeI: Would also prefer this to be a var instead of hardcoded if above comment is wrong ^

SequeI:
Additionally, ideally you could add a workable test for this within `user_provided` just to ensure this new feature works

molecule/user_provided/vars/vars.yml
JasonPowr: This was added here because the molecule tests were failing with this error message 
```
The conditional check 'tas_single_node_rekor_redis.backfill_redis.enabled' failed. The error was: error while evaluating conditional (tas_single_node_rekor_redis.backfill_redis.enabled): 'dict object' has no attribute 'backfill_redis'
```

I believe because backfill_redis is a property under tas_single_node_rekor_redis, it needed to be defined here aswell.

roles/tas_single_node/templates/manifests/rekor/backfill_redis.sh.j2
JasonPowr: The file paths here are correct, {{ tas_single_node_certs_dir }}/ingress-rekor.pem refers to the file path in the rhel machine which is mounted inside the bacfill_redis pod to the /usr/local/share/ca-certificates/ingress-rekor.pem file using the -v (volume) flag 

eg     -v "(file to mount):(path to mount it to)" \

I think that variables here dont really make sense as /usr/local/share/ca-certificates is the default for most machines when it comes to ca certificates

JasonPowr:
> Additionally, ideally you could add a workable test for this within `user_provided` just to ensure this new feature works

Im not really sure there is a good way to test this if I am honest, ill take a look, but really we would only be testing to make sure that we can configure it with ansible and thats already done here

molecule/user_provided/vars/vars.yml
SequeI: Ah yes, I didn't see you had put the new job config under `tas_single_node_rekor_redis`

roles/tas_single_node/templates/manifests/rekor/backfill_redis.sh.j2
```diff
@@ -0,0 +1,38 @@
...
+    --add-host="rekor.{{ tas_single_node_base_hostname }}:{{ dns_resolver }}" \
```
bkabrda, line 26: I don't think the dns_resolver/add-host is necessary at all. All pods on the same network (use --network) should be able to connect to each other - see e.g. the rekor-server.j2 to see how it connects to trillian and Redis.

roles/tas_single_node/templates/manifests/rekor/backfill_redis.sh.j2
```diff
@@ -0,0 +1,38 @@
...
+else
+  startIndex=0
+fi
+
+if [ "$startIndex" -gt "$endIndex" ]; then
+  startIndex=0
+fi
+
+if [ "$startIndex" -le "$endIndex" ]; then
+  podman run --rm --name backfill_redis \
```
bkabrda, line 23: Can we use `--force` instead of `--rm`? That should ensure that the previous container always stays on the node (until the cron job runs next time, which will replace it), so if there's a failure, the user has time to capture logs.

roles/tas_single_node/README.md
```diff
@@ -18,7 +18,7 @@ Deploy the [RHTAS](https://docs.redhat.com/en/documentation/red_hat_trusted_arti
 |Option|Description|Type|Default|
 |---|---|---|---|
 | tas_single_node_podman_network | Name of the Podman network for containers to use. | str |  `rhtas`  |
-| tas_single_node_rekor_redis | Details on the Redis connection for Rekor. You can set this to a custom Redis instance. | dict of 'tas_single_node_rekor_redis' options |  `{'database_deploy': True, 'redis': {'host': 'rekor-redis-pod', 'port': 6379, 'password': 'password'}}`  |
+| tas_single_node_rekor_redis | Details on the Redis connection for Rekor. You can set this to a custom Redis instance. | dict of 'tas_single_node_rekor_redis' options |  `{'database_deploy': True, 'redis': {'host': 'rekor-redis-pod', 'port': 6379, 'password': 'password'}, 'backfill_redis': {'enabled': True, 'schedule': {'minute': '0', 'hour': '0', 'day': '*', 'month': '*', 'weekday': '*'}}}`  |
```
bkabrda, line 21: TBH I think this shouldn't live under the `tas_single_node_rekor_redis` key, but be its own key. The `tas_single_node_rekor_redis` key is supposed to be for redis connection and this kinda stretches it (and, as you see yourself in the user_provided scenario change, once you provide `tas_single_node_rekor_redis`, you have to define the backfill job parameters, which is not great)

roles/tas_single_node/templates/manifests/rekor/backfill_redis.sh.j2
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
```
bkabrda, line 1: While I provided some comments throughout this file, I think a better (although much laborious) route to take would be:
* Create a systemd timer unit
* Create a k8s manifest and use our standard approach to create a systemd service unit for it to run
* Make the timer unit run the job on schedule

roles/tas_single_node/templates/manifests/rekor/backfill_redis.sh.j2
```diff
@@ -0,0 +1,38 @@
...
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
+LAST_INDEX_FILE="/tmp/rekor_last_index"
```
bkabrda, line 4: Is the last index not queryable? Storing in a `/tmp` file is not great, because that will get wiped out on reboot. Can we either make this dynamic (query for it) or figure out a more reliable approach to store it (would a podman secret work? or at least a better file location somewhere in /var?)

bkabrda (changes requested):
:wave: thanks for the PR Jason. I think this is a very good start, but I put some suggestions inline to improve some of the mechanics of this functionality.

roles/tas_single_node/templates/manifests/rekor/backfill_redis.sh.j2
JasonPowr: I did try this first, I didn't have much luck if I am honest, backfill redis is awkward when connecting to the rekor instance it has to be a url so it can query the /api/v1/log/entries?logIndex=x api endpoint 

I will take another look, also when I think about it the unable to connect may have been related to the issue fixed in https://github.com/securesign/artifact-signer-ansible/pull/100, either way I will take a look

roles/tas_single_node/templates/manifests/rekor/backfill_redis.sh.j2
JasonPowr: Im not familiar with systemd timer unit, so i will take a look :) thanks for the feedback

roles/tas_single_node/templates/manifests/rekor/backfill_redis.sh.j2
JasonPowr: AFAIK, the last index filled in is not queryable, I will take a look at a better way to store it

bkabrda (changes requested):
Hi, thanks for putting all the additional work in. This looks really good, I just have one minor comment on the config variable we're introducing.

Also a question: Did you try to see if the timer execution works correctly? (e.g. by modifying the timer locally to execute every minute or something like that to actually see it execute repeatedly)

Thinking about this further, I think we should expose the timer configuration as a variable under `tas_single_node_backfill_redis` (see my first comment about modifying this variable). This will both allow users to run the job when they want to and allow us to easily test this feature.

JasonPowr:
> Also a question: Did you try to see if the timer execution works correctly? (e.g. by modifying the timer locally to execute every minute or something like that to actually see it execute repeatedly)

I did yes, everything worked as expected :) 

> I think we should expose the timer configuration as a variable under tas_single_node_backfill_redis

Makes sense :) sure thing

roles/tas_single_node/templates/manifests/rekor/backfill_redis.j2
```diff
@@ -0,0 +1,68 @@
...
+                --rekor-address=http://{{ tas_single_node_rekor_server_pod }}-pod:{{ tas_single_node_rekor_server_port_http }} \
+                --start="${startIndex}" --end="${endIndex}"
+
+              redis-cli \
+                -h {{ tas_single_node_rekor_redis.redis.host }} \
+                -p {{ tas_single_node_rekor_redis.redis.port }} \
+{% if tas_single_node_rekor_redis.redis.password != "" %}
+                -a "{{ tas_single_node_rekor_redis.redis.password }}" \
+{% endif %}
+                SET last_filled_index "$((endIndex + 1))"
```
bkabrda, line 68: Just a minor question here: if for whatever reason this last command fails and doesn't update the `last_filled_index`, what will happen when we execute next time? Will we have duplicated data or will it just take longer to execute but all the data will be correct in the end?

bkabrda (approved):
LGTM, great job! Thanks.

PR #103, opened 2024-11-13
Adding user supplied CA for custom OIDC providers

SequeI:
Adding the ability to use your own CA cert for custom OIDC providers

From my understanding of the ticket, This PR can do the following
* User can supply their own CA cert which will be placed within /certs and used to validate the auth of a custom OIDC provider by checking against the signed ID token and Access token issued by the provider.

I placed the cert within /certs as I heard /trusted-ca is used for TLS certs and this isn't TLS but just identity auth from what I understand

lmk if you prefer different var names or if you want the cert location to be /trusted-ca same with Trillian CA. I will create a test in `user_provided` just to check if the cert is being consumed correctly and placed within /certs. Not sure if there is a manual option to configure to check CA against an OIDC or it's configured to try automatically, but will also set if possible.

### DIFF
--- a/roles/tas_single_node/tasks/podman/fulcio.yml
+++ b/roles/tas_single_node/tasks/podman/fulcio.yml
@@ -34,24 +34,10 @@
       data:
         config.yaml: |
           {{ fulcio_config_content }}
-  register: configmap_result
 
-- name: Create Fulcio Trusted OIDC CA Configmap
-  ansible.builtin.copy:
-    content: "{{ configmap_content | to_nice_yaml(indent=2) }}"
-    dest: "{{ tas_single_node_fulcio_trusted_oidc_ca_config }}"
-    mode: "0600"
-  vars:
-    configmap_content:
-      kind: ConfigMap
-      apiVersion: v1
-      metadata:
-        name: "{{ tas_single_node_fulcio_trusted_oidc_ca_configmap_name }}"
-        namespace: fulcio-system
-      data:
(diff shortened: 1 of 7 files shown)


tommyd450:
@SequeI Changes seem reasonable to me but you should create some sort of steps or example of configuring using Keycloak or equivalent.

roles/tas_single_node/tasks/podman/fulcio.yml
```diff
@@ -36,6 +36,23 @@
           {{ fulcio_config_content }}
   register: configmap_result
 
+- name: Create Fulcio Trusted OIDC CA Configmap
```
bkabrda, line 39: Instead of creating multiple configmaps, would it be possible to have just a single one and use subPath? (E.g. see https://stackoverflow.com/a/72327498)

bkabrda:
I think this looks really good, I just let one minor comment inline; I also wanted to ask you whether you actually tested that this works. Thanks!

roles/tas_single_node/tasks/podman/fulcio.yml
SequeI: No idea this existed, thanks!

SequeI:
@bkabrda Tested, correctly get's consumed and formatted into a .pem file at /certs within the fulcio-server container alongside the other ingress certs. SSL ENV already instructs the golang crypto/tls package to look at this custom cert dir.

bkabrda (approved):
LGTM, thank you!

PR #105, opened 2024-11-14
[SECURESIGN-1465] Enable customer provided cert for Fulcio

fghanmi:
Changes in this PR:
- Accept user-provided certificate/private key (if not, generate automatically Fulcio root - as before)
- Allow the configuration of the Certificate attributes {organization_name, organization_email, common_name}


bkabrda (changes requested):
Thanks for the PR, I think this works really well, except some scenarios with cert/key changes that I think would not be covered right now. Specifically, in the `certificates.yml` file:
* The whole `Create Fulcio root` block is only ran if one of the files is missing on the managed node. We can no longer do this, as we have to expect the user to e.g. first let the role generate key automatically and then later provide their own cert or change the settings somehow (or rotate their provided cert etc). I think this will be tricky to handle but we have to do it and consider all scenarios when user changes from one way of doing something to another way.
* The same applies to the `Create Fulcio public key` task, which would now only run when the public key file doesn't exist (because of the `creates`).
* Same for `Create self-signed Fulcio root from CSR`.
* I think the `Create certificate signing request (CSR) for Fulcio root certificate` with your changes will run every time and generate a different cert, which is not what we want.

I know this is going to be tedious, but we have to switch from the "do these things" mindset here to "ensure the files are in the desired state". Let me know if you need any help.

roles/tas_single_node/README.md
```diff
@@ -20,6 +20,9 @@ Deploy the [RHTAS](https://docs.redhat.com/en/documentation/red_hat_trusted_arti
...
 | tas_single_node_rekor_redis | Details on the Redis connection for Rekor. You can set this to a custom Redis instance. | dict of 'tas_single_node_rekor_redis' options |  `{'database_deploy': True, 'redis': {'host': 'rekor-redis-pod', 'port': 6379, 'password': 'password'}}`  |
 | tas_single_node_trillian | Details on the database connection for Trillian. You can set this to a custom MySQL or MariaDB instance. | dict of 'tas_single_node_trillian' options |  `{'database_deploy': True, 'mysql': {'user': 'mysql', 'root_password': 'rootpassword', 'password': 'password', 'database': 'trillian', 'host': 'trillian-mysql-pod', 'port': 3306}}`  |
+| tas_single_node_fulcio | Details on the certificate settings for Fulcio. You can configure the certificate with the organization's details. | dict of 'tas_single_node_fulcio' options |  `{'certificate': {'organization_name': '', 'organization_email': '', 'common_name': ''}}`  |
+| tas_single_node_fulcio_private_key | Private key for Fulcio, used for signing root certificate. | str |  |
+| tas_single_node_fulcio_root_ca | The root certificate for Fulcio. | str |  |
```
bkabrda, line 25: Should this be mutually exclusive with `tas_single_node_fulcio`? If so, then let's mention it here and also create a task that would validate that only one of these is provided.

bkabrda: Also, maybe we should also nest `tas_single_node_fulcio_private_key` and `tas_single_node_fulcio_root_ca` under `tas_single_node_fulcio` (on the same level as the `certificate` attribute). I think it would make more sense to set these values at a single place.

roles/tas_single_node/tasks/certificates.yml
```diff
@@ -93,6 +88,16 @@
...
+      when: >
+        tas_single_node_fulcio_private_key == ""
+        and (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_fulcio_private_key) | list | length) == 0
+
+    - name: Load user-provided Fulcio private key
+      ansible.builtin.copy:
+        src: "{{ tas_single_node_fulcio_private_key }}"
+        dest: "{{ tas_single_node_remote_fulcio_private_key }}"
+        mode: '0600'
+      when: tas_single_node_fulcio_private_key != ""
```
bkabrda, line 100: For the two private key tasks, I think we should:
* Use a different path for the key file for each task. For example, if the user first provided a key themselves but later wanted to switch to a role-generated key, we would just happily keep using their provided old key. Using different paths for the file should solve this.
* For the user-provided key, I think if it's not provided we should explicitly ensure that the file is absent (we don't want users credentials laying around on the node unnecessarily).
* Based on what the user configured, we can then copy either the role generated key or the user key to `tas_single_node_remote_fulcio_private_key` as the final destination (so we won't have to change any tasks that depend on that file location).

roles/tas_single_node/tasks/certificates.yml
```diff
@@ -77,11 +77,6 @@
...
 
 - name: Create Fulcio root
-  when: >
-    (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_fulcio_private_key) | list | length) == 0
-    or (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_fulcio_public_key) | list | length) == 0
-    or (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_fulcio_root_ca) | list | length) == 0
-
   block:
     - name: Create Fulcio private key
       # NOTE: As of cosign 2.3.0, only RSA with > 2048 B pkey size or ECC with prime256v1 (==secp256r1) are supported
```
bkabrda, line 82: I think based on this comment we should document for users what type of key can be provided.

roles/tas_single_node/tasks/certificates.yml
```diff
@@ -101,18 +106,30 @@
           -in '{{ tas_single_node_remote_fulcio_private_key }}'
           -out '{{ tas_single_node_remote_fulcio_public_key }}'
           -passin 'pass:{{ tas_single_node_fulcio_ca_passphrase }}'
-        creates: "{{ tas_single_node_remote_fulcio_public_key }}"
+      changed_when: >
+        (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_fulcio_public_key) | list | length) == 0
+      when: >
+        tas_single_node_fulcio_private_key != ""
+        or (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_fulcio_public_key) | list | length) == 0
```
bkabrda, line 113: * Will this command work even if user provides an RSA key?
* For RSA keys, generating the public key can yield a different output every time. I think what we should do is we should have a task that finds out whether the public key corresponds to the private key; if not, then this task should regenerate.

roles/tas_single_node/tasks/certificates.yml
```diff
@@ -126,6 +143,17 @@
           -passin 'pass:{{ tas_single_node_fulcio_ca_passphrase }}'
           -out '{{ tas_single_node_remote_fulcio_root_ca }}'
         creates: "{{ tas_single_node_remote_fulcio_root_ca }}"
+      when: >
+        tas_single_node_fulcio_root_ca == ""
+        and (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_fulcio_root_ca) | list | length) == 0
+        or tas_single_node_fulcio_private_key != ""
```
bkabrda, line 149: Similar to one of the previous comments - if user first provided cert on their own and then they wanted to switch to role-generated cert, this wouldn't properly regenerate. I think the last line of the conditon should be replaced by checking a result of a new task which will verify whether the existing private/public key match the existing cert.

bkabrda (changes requested):
Hi Firas, thanks for implementing the requested changes. I think this version is better, but there are still some scenarios that need to be covered - see the comments inline.

roles/tas_single_node/defaults/main.yml
```diff
@@ -51,6 +59,9 @@ tas_single_node_tsa_ca_passphrase: rhtas
 tas_single_node_tsa_signer_passphrase: rhtas
 tas_single_node_ct_logprefix: rhtasansible
 
+tas_single_node_fulcio_private_key: ""
+tas_single_node_fulcio_root_ca: ""
```
bkabrda, line 63: We now have the nested values under `tas_single_node_fulcio`, so we can remove these ones, right?

roles/tas_single_node/README.md
```diff
@@ -79,6 +82,22 @@ Deploy the [RHTAS](https://docs.redhat.com/en/documentation/red_hat_trusted_arti
 | root_password | The root password for the database. | str | no |  |
 | database | The database name to connect to. | str | no |  |
 
+#### Options for main > tas_single_node_fulcio
+
+|Option|Description|Type|Required|Default|
+|---|---|---|---|---|
+| certificate | Details on the certificate attributes for Fulcio. | dict of 'certificate' options | no |  |
+| tas_single_node_fulcio_private_key | The user-provided private key for Fulcio, used for signing root certificate (only RSA with > 2048 B pkey size or ECC with prime256v1 (==secp256r1) are supported). | str | no |  |
+| tas_single_node_fulcio_root_ca | The user-provided root certificate for Fulcio. | str | no |  |
```
bkabrda, line 91: Let's mention that this is mutually exclusive with `certificate` (it is, right?)

roles/tas_single_node/tasks/certificates.yml
```diff
@@ -90,9 +85,69 @@
...
+        path: "{{ tas_single_node_remote_fulcio_user_provided_private_key }}"
+      register: user_provided_key_stat
+
+    - name: Copy auto-generated Fulcio private key into tas_single_node_remote_fulcio_private_key
+      ansible.builtin.copy:
+        src: "{{ tas_single_node_remote_fulcio_auto_generated_private_key }}"
+        dest: "{{ tas_single_node_remote_fulcio_private_key }}"
+        mode: '0600'
+        remote_src: true
+      when: auto_generated_key_stat.stat.exists
```
bkabrda, line 131: I don't think we need to do these 2 stats calls. We can use the user-provided key when `tas_single_node_remote_fulcio_user_provided_private_key` is defined and non-empty, otherwise use the auto-generated key. Does that make sense?

roles/tas_single_node/tasks/certificates.yml
```diff
@@ -90,9 +85,69 @@
...
+        dest: "{{ tas_single_node_remote_fulcio_private_key }}"
+        mode: '0600'
+        remote_src: true
+      when: user_provided_key_stat.stat.exists
+
+    - name: Verify if public key matches the private key
+      ansible.builtin.shell:
+        cmd: >-
+          set -o pipefail &&
+          openssl ec -pubout -in '{{ tas_single_node_remote_fulcio_private_key }}' -passin 'pass:{{ tas_single_node_fulcio_ca_passphrase }}' |
```
bkabrda, line 145: Will this work if the private key is RSA?

bkabrda (changes requested):
Thanks for putting all the work into this, it looks much better now. I left couple comments inline, but they're mostly minor at this point (still need to be addressed though).

roles/tas_single_node/tasks/certificates.yml
bkabrda: Heh, interesting, I didn't expect that it would :)

roles/tas_single_node/meta/argument_specs.yml
```diff
@@ -118,6 +118,67 @@ argument_specs:
...
+            type: "str"
+            required: false
+            version_added: "1.1.1"
+          tas_single_node_fulcio_root_ca:
+            description: "The user-provided root certificate for Fulcio. This field is mutually exclusive with the certificate field."
+            type: "str"
+            required: false
+            version_added: "1.1.1"
+
+      tas_single_node_fulcio_private_key:
```
bkabrda, line 168: You removed this one (and the one below) from the defaults/main.yml file, but not from here, so they're still rendered in the README. Let's remove them from here also.

bkabrda (changes requested):
:wave: thanks for the additional fixes. There is still one minor thing to do (remove the 2 no longer existing variables from arguments spec) and then I can approve. Nice work!

bkabrda (approved):
LGTM now, thanks for putting in all the work!

PR #111, opened 2024-11-29
Adding Dex as a mock OIDC 

SequeI:
Added mock OIDC server + Identity provider, get's rid of our dependency on an external uncontrolled keycloak instance.

### DIFF
--- a/molecule/dex-config.yaml
+++ b/molecule/dex-config.yaml
@@ -1,0 +1,31 @@
+---
+- name: Podman
+  ansible.builtin.package:
+    name: podman
+    state: present
+
+- name: Create dir
+  ansible.builtin.file:
+    path: "/etc/dex/config"
+    state: directory
+    mode: "0755"
+
+- name: Copy Dex config over
+  ansible.builtin.copy:
+    src: "../../vm-testing/dex-config.yaml"
+    dest: "/etc/dex/config/dex.yaml"
+    mode: "0644"
+
+- name: Run Dex container
+  ansible.builtin.shell: |
+      sudo podman run -d \
+      --name dex-idp \
+      -p 5556:5556 \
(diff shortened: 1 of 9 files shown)


molecule/dex-config.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+- name: Copy Dex config over
+  ansible.builtin.copy:
+    src: "../../vm-testing/dex-config.yaml"
+    dest: "/etc/dex/config/dex.yaml"
+    mode: "0644"
+
+- name: Run Dex container
+  ansible.builtin.shell: |
+      sudo podman run -d \
```
bkabrda, line 21: If we need the sudo here, let's do it the ansible way with become/become_user.

molecule/dex-config.yaml
```diff
@@ -0,0 +1,31 @@
+---
+- name: Podman
+  ansible.builtin.package:
+    name: podman
+    state: present
+
+- name: Create dir
```
bkabrda, line 7: Let's use more descriptive task names, e.g. "Create Dex config directory"

vm-testing/test/test-sign-blob.sh
```diff
@@ -13,31 +13,45 @@ cp "${BASE_HOSTNAME}".pem /etc/pki/ca-trust/source/anchors/
 update-ca-trust
 
 # set up cosign env
-export KEYCLOAK_REALM=trusted-artifact-signer
+export OIDC_ISSUER_URL=http://dex-idp:5556/dex
```
bkabrda, line 16: Let's make the OIDC_ISSUER_URL come from outside of the script to make it a little more flexible.

bkabrda (approved):
Thanks! LGTM.

PR #114, opened 2024-12-03
Adding testing resources

SequeI:
Adding checks for new Dex playbook so you are able to rerun it with no errors, plus a simple check to verify.yml so you can also rerun it.

Adding testing resources so people without TF access can easily deploy the collection and test

### DIFF
--- a/molecule/default/verify.yml
+++ b/molecule/default/verify.yml
@@ -7,9 +7,26 @@
   vars_files:
     - vars/vars.yml
   tasks:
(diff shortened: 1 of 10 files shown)


SequeI:
Disconnecting dex-idp from podman network, found an issue with network flapping causing the container to be unreachable after max one attempt, which is why it worked so far in CI but would fail for local development if someone tried to resign/reverify

tommyd450:
@SequeI add some notes for this to the read me, otherwise its all good to go.

PR #116, opened 2024-12-11
Adding aws to CI tests

SequeI:
Adding AWS EC2 to our pipeline
Spins up m.5 instance and runs `molecule test`

### DIFF
--- a/molecule/default/create.yml
+++ b/molecule/default/create.yml
@@ -24,10 +24,6 @@
           "tag:Name": CI-TEST
       register: vpcs
 
-    - name: print
-      debug:
-        msg: "{{ vpcs }}"
-
     - name: Create VPC if it does not exist yet
       amazon.aws.ec2_vpc_net:
         name: "CI-TEST"
@@ -36,10 +32,6 @@
         state: present
       when: vpcs.vpcs | length == 0
       register: created_vpc
-
-    - name: print
-      debug:
-        msg: "{{ created_vpc }}"
 
     - name: Set VPC ID if new VPC created
       ansible.builtin.set_fact:
@@ -58,10 +50,6 @@
           vpc-id: "{{ vpc_id }}"
           "tag:Name": CI-TEST
       register: subnets
-
-    - name: print
-      debug:
-        msg: "{{ subnets }}"
 
     - name: Create subnet if it does not exist
       amazon.aws.ec2_vpc_subnet:
@@ -91,10 +79,6 @@
           "tag:Name": "CI-TEST"
       register: igw_facts
 
-    - name: print
-      debug:
-        msg: "{{ igw_facts }}"
-
     - name: Create internet gateway if it does not exist
       amazon.aws.ec2_vpc_igw:
         vpc_id: "{{ vpc_id }}"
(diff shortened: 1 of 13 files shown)


SequeI:
Pipeline works - molecule/README.md needs work to configure for local development workflow
Take a look at benchmarks to see if an instance with more cores is faster/does it make sense

tommyd450:
I honestly cannot see anything that sticks out as wrong code wise. Structure wise do you think it might be worth splitting it out from the create files and dumping them as a task within the molecule/default/tasks/ directory.
@bkabrda @SequeI wdyt?

Outside of that it lgtm

testing-requirements.txt
```diff
@@ -1,5 +1,8 @@
 ansible-lint==24.9.0
 antsibull-docs==2.16.1
+awscli==1.36.19
+botocore==1.35.78
+boto3==1.35.78
```
bkabrda, line 5: Let's remove the tft-cli dependency from this file, we'll no longer need it.

molecule/default/vars/vars.yml
```diff
@@ -8,3 +8,7 @@ tas_single_node_cockpit:
   enabled: false
 tas_single_node_registry_username: "{{ lookup('env', 'TAS_SINGLE_NODE_REGISTRY_USERNAME') }}"
 tas_single_node_registry_password: "{{ lookup('env', 'TAS_SINGLE_NODE_REGISTRY_PASSWORD') }}"
+private_key_file_path: ~/.ssh/id_rsa
+image_id: ami-0cd93da2da7f10cab
+instance_type: m5.xlarge
+aws_key_name: CI-TEST
```
bkabrda, line 14: These variables should go to `molecule/default/molecule.yml` to the `platforms` config (and then you can access them via `molecule_yml.platforms`)

molecule/default/create.yml
```diff
@@ -15,94 +15,194 @@
...
-        instance_conf: "{{ instance_config_dict.results | map(attribute='ansible_facts.instance_conf_dict') | list }}"
-
-    - name: Debug instance_conf
-      ansible.builtin.debug:
-        msg: "{{ instance_conf }}"
-      when: debug_outputs
+        instanceID: "{{ ec2_instance.instances[0].instance_id}}"
+        instanceIP: "{{ ec2_instance.instances[0].network_interfaces[0].association.public_ip }}"
 
-    - name: Dump instance config
```
bkabrda, line 98: I think it would be better to keep using the dumped instance config - that provides a way for molecule to understand whether the VM already exists or not. If we're ever going to use this locally for development, running `molecule create` twice wouldn't actually spawn 2 VMs, it would know that the VM is already spawned and inform the user of it.

You can also save the instance ID in the molecule instance config to access it in `destroy.yml`.

SequeI:
@tommyd450 

I fully agree, this will be good to do I think during the ticket to make the 2 scenarios run in parallel while network infra instantiates before they are run. For now it is probably fine as we get used to this new system but good idea!

bkabrda (approved):
LGTM now, this is awesome. Thanks!

PR #119, opened 2024-12-23
[SECURESIGN-1659] Add AAP testing scenario for collection

fghanmi:
Before adding schedule cron: 
last aap test: https://github.com/securesign/artifact-signer-ansible/actions/runs/13284566452/job/37090680390


.github/workflows/aap.yaml
```diff
@@ -0,0 +1,53 @@
+name: Using Ansible Molecule, test RHTAS collection from a private AAP
+
+# on:
```
fghanmi, line 3: To be uncommented once the PR is reviewed and build log is verified.

tommyd450:
LGTM we can monitor it over the coming weeks to see if everything is up to scratch!

SequeI:
Thanks for the PR Firas! It looks in really solid shape 

@bkabrda I believe we could symlink a bunch of the unchanged files, or put some of the generic tasks like standing up the AWS infra into their own yml files and just `include_tasks` them to reduce code duplicity, WDYT? Although not sure if this is a big deal since this is just for testing.

SequeI (changes requested):
Very good, just left a comment for Slavek to take a look at that might make sense for this PR

Thanks Firas!

molecule/aap-setup/converge.yml
```diff
@@ -0,0 +1,128 @@
+---
+- name: Converge
+  hosts: molecule
+  gather_facts: true
+  tags: aap_installation
+  vars_files:
+    - vars/vars.yml
+    - vars/aap_nodes.yml
+  tasks:
+    - name: Error out if registry username is not set
```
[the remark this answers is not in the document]
fghanmi: @SequeI , I don't think we can use it, the when condition is based on different variable here: `aap_setup_prep_inv_secrets.all.registry_username`

molecule/aap-collection-test/tasks/_parse_tf_results.yml
```diff
@@ -0,0 +1,12 @@
+---
```
bkabrda, line 1: This file shouldn't be necessary as we're not using TF anymore.

bkabrda:
I definitely agree with @SequeI that we should reuse as much code as possible here by symlinking and/or reusing Ansible task files. Can you do that before I do a full review @fghanmi? Thanks!

.github/workflows/aap.yaml
```diff
@@ -0,0 +1,54 @@
...
+            AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+            AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+            GITHUB_RUN_ID: ${{ github.run_id }}
+            REDHAT_API_TOKEN: ${{ secrets.REDHAT_API_TOKEN }}
+          run: |
+            source venv/bin/activate
+            molecule reset
+            molecule -v create --scenario-name aap-setup
+            molecule -v converge --scenario-name aap-setup
+            molecule -v test --scenario-name aap-collection-test
```
bkabrda, line 44: I think this isn't using Molecule as it was meant to be used. This whole "setup" should be a part of the `prepare.yml` file inside the same Molecule scenario.

Furthermore, we shouldn't install the collection from AAP and run it locally, we want AAP to use it. Let me explain in more detail:

* Have 1 Molecule testing scenario, which creates 2 VMs
* In `prepare.yml`:
  * Install AAP on one of the VMs
  * Upload the collection to AAP
  * Create all AAP objects necessary to run the collection from AAP (something similar to https://docs.redhat.com/en/documentation/red_hat_ansible_automation_platform/2.4/html/developing_ansible_automation_content/publishing-playbook-collection-aap#create-aap-job_publishing-playbook-collection-aap-intro)
* In `converge.yml`:
  * Trigger the AAP job and poll until it finishes
* In `verify.yml`:
  * Verify that TAS was installed correctly (e.g. use the verify from the `default` Molecule scenario)

Does that make sense? The important point here is that we need to test not only uploading the collection to AAP, but we need to ensure that it executes correctly *from AAP*.

molecule/aap-test/verify.yml
```diff
@@ -0,0 +1,88 @@
+- name: Verify
+  hosts: worker_group
+  gather_facts: true
+  become: true
+  vars:
+    test_dir_path: "{{ lookup('env', 'MOLECULE_SCENARIO_DIRECTORY') }}/../../vm-testing/test"
+  vars_files:
+    - vars/vars.yml
+  tasks:
```
bkabrda, line 9: Most of this file seems to be C&P from https://github.com/securesign/artifact-signer-ansible/blob/main/molecule/default/verify.yml - can we pull the shared tasks out to a common shared file?

molecule/aap-test/molecule.yml
```diff
@@ -0,0 +1,24 @@
...
+dependency:
+  name: galaxy
+  options:
+    requirements-file: molecule/requirements.yml
+driver:
+  name: default
+platforms:
+  - name: instance
+    private_key_file_path: ~/.ssh/id_rsa
+    image_id: ami-0cd93da2da7f10cab
```
bkabrda, line 11: Let's also use `ami-0dac29b55bd7c33aa` throughout this PR please (RHEL 9.4, upgraded for the whole collection in https://github.com/securesign/artifact-signer-ansible/pull/139/files)

molecule/aap-test/converge.yml
```diff
@@ -0,0 +1,368 @@
...
+            url: "https://www.redhat.com"
+        state: present
+        ah_host: https://{{ aap_instance_ip }}/
+        ah_token: "{{ ah_token.token }}"
+        validate_certs: false
+
+    - name: Switch images from `registry.redhat.io` images to the dev images
+      delegate_to: localhost
+      ansible.builtin.shell: |
+        sed -f ../../vm-testing/dev-images.sed  -i ../../roles/tas_single_node/defaults/main.yml
```
SequeI, line 132: I believe you could get rid of this task and reference Tomas' image mirroring PR where he adds the `podman.yml` as a `vars_files` and symlink the `podman.yml` in this scenarios `vars` folder [Like this](https://github.com/securesign/artifact-signer-ansible/pull/89/files#diff-bc4d0c4551c2773138cb727ed62a97e7e37f37731b8f6a2c8fe715ae136fdfb6R7) and [this](https://github.com/securesign/artifact-signer-ansible/pull/89/files#diff-521f50ba7612c8c6b66607e8d5a15c1890b8bc01bdab29beb55f814f54edf1f1). lmk what you think, might be better to reuse what is used in other scenarios currently

molecule/aap-test/converge.yml
```diff
@@ -0,0 +1,368 @@
...
+                  aap_instance_ip: "{{ aap_instance_ip }}"
+                  token: "{{ ah_token.token }}"
+                  tas_single_node_oidc_issuers:
+                    - issuer: "http://dex-idp:5556/dex"
+                      url: "http://dex-idp:5556/dex"
+                      client_id: example-app
+                      type: email
+                  tas_single_node_base_hostname: myrhtas
+                  tas_single_node_cockpit:
+                    enabled: false
```
SequeI, line 349: This config is false by default so not sure it is needed here

molecule/aap-test/create.yml
```diff
@@ -0,0 +1,153 @@
...
+          {{ instances }}
+
+    - name: Populate instance config dict
+      ansible.builtin.set_fact:
+        instance_conf_dict: {
+          'instance': "{{ item.value.id }}",
+          'address': "{{ item.value.ip }}",
+          'user': "ec2-user",
+          'port': "22",
+          'identity_file': "{{ molecule_yml.platforms[0].aws_key_name }}",
```
SequeI, line 72: Might be worthwhile creating a separate AAP-KEY keypair on AWS just to differentiate between the regular scenario CI runs and this scenario since it's a special midnight workflow. You can just put is in as an env var the same way it is shown in the [README](https://github.com/securesign/artifact-signer-ansible/blob/main/molecule/README.md)

SequeI (approved):
/lgtm left some very small comments, but this is essentially finished from what I have seen so far. Thanks a lot for dividing up the common tasks in their own files and editing the existing scenarios with them! Thanks Firas, this was especially frustrating to work on for sure

molecule/aap-test/create.yml
fghanmi: @SequeI Are there any specific gains or reasons for creating a separate AAP-KEY keypair for this ? Just wondering if it provides any concrete benefits beyond differentiation

SequeI: @fghanmi Purely differentiation, esp as our collection grows mature and more complex in it's testing suite, might just make things easier in the future. No big deal though, up to you if you think it's worth it

SequeI (approved):
/lgtm Great, thanks Firas! 💯

bkabrda (approved):
Great job on this Firas, this is massive (and massively great and useful!)

PR #127, opened 2025-01-23
Add e2e test scenario

bouskaJ:
This PR reuses existing installation that was provided by molecule framework and add execution of https://github.com/securesign/sigstore-e2e test suite

### DIFF
--- a/.github/actions/go-e2e/action.yaml
+++ b/.github/actions/go-e2e/action.yaml
@@ -1,0 +1,88 @@
+name: 'Go E2E Tests Reusable Action'
+description: 'Run Go E2E tests using the provided TAS server and Go version.'
+inputs:
+  tas_address:
+    description: 'TAS server address'
+    required: true
+    type: string
+  fulcio_host:
+    description: 'Fulcio server hostname'
+    required: false
+    default: 'fulcio.myrhtas'
+  rekor_host:
+    description: 'Rekor server hostname'
+    required: false
+    default: 'rekor.myrhtas'
+  rekor_search_host:
+    description: 'Rekor search server hostname'
+    required: false
+    default: 'rekor-search.myrhtas'
+  tuf_host:
+    description: 'TUF server hostname'
+    required: false
+    default: 'tuf.myrhtas'
+  tsa_hostname:
+    description: 'TSA server url'
+    required: false
+    default: 'tsa.myrhtas'
+  cli_server_host:
+    description: 'Client server hostname'
+    required: false
+    default: 'cli-server.myrhtas'
+  oidc_host:
+    description: 'OIDC provider hostname'
+    required: false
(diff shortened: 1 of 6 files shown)


.github/actions/go-e2e/action.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+    - name: Resolve TAS_SINGLENODE hostname
+      shell: bash
+      run: |
+        echo "${{ inputs.tas_address }} fulcio.myrhtas tuf.myrhtas rekor.myrhtas tsa.myrhtas rekor-search.myrhtas cli-server.myrhtas dex-idp" | sudo tee -a /etc/hosts
+
+    - name: Resolve certificates
+      shell: bash
+      run: |
+        openssl s_client -showcerts -connect rekor.myrhtas:443 < /dev/null | awk '/BEGIN CERTIFICATE/,/END CERTIFICATE/{ if(/BEGIN CERTIFICATE/){a++}; out="cert"a".pem"; print >out}'
```
bkabrda, line 34: We do a lot of the same stuff as you do here in https://github.com/securesign/artifact-signer-ansible/blob/main/vm-testing/test/test-sign-blob.sh#L7 - could you possibly try to extract the common parts of the referenced file and your script to a common shared file and reuse it here? It'd be nice to prevent the code duplication.

bouskaJ: Good point! Finally, after some suffering time, I was able to reuse the same function in both scenarios.

vm-testing/inventory-sample
```diff
@@ -1,7 +0,0 @@
-[rhtas]
```
bkabrda, line 1: What is the reason for removing this file in this PR?

.github/actions/go-e2e/action.yaml
```diff
@@ -0,0 +1,50 @@
...
+        path: e2e
+    - name: Install Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: ${{ inputs.go_version }}
+
+    - name: Resolve TAS_SINGLENODE hostname
+      shell: bash
+      run: |
+        echo "${{ inputs.tas_address }} fulcio.myrhtas tuf.myrhtas rekor.myrhtas tsa.myrhtas rekor-search.myrhtas cli-server.myrhtas dex-idp" | sudo tee -a /etc/hosts
```
bkabrda, line 29: Nit: for this step and for the last one, could we have variables defining the service hostnames? We're hardcoding a bunch of values in two places.

vm-testing/inventory-sample
bouskaJ: :anguished: mistake. Thank you for spotting this.

bkabrda (approved):

PR #132, opened 2025-01-28
Rename and fix up port variables

bkabrda:
This PR renames some variables related to ports and fixes their usage. This still means that all container ports are exposed as host ports.

In a followup PR to this one, I will bump up the requirement to RHEL 9.4 (podman >= 4.9) which makes it possible to explicitly select ports to expose and will make sure the hostPort definitions are only done on the proper ports.


SequeI (approved):
/lgtm

PR #135, opened 2025-01-31
Tune Molecule settings to speed up CI runs

bkabrda:
This PR introduces bunch of options that significantly speed up Molecule CI runs; by my local testing, this will shave off more than five minutes of every molecule run (speeding up the first converge by ~3:30 and the idempotence converge by ~2:00).


SequeI (approved):
/lgtm 25m -> 13mins is crazy

PR #139, opened 2025-02-05
Require RHEL >= 9.4

bkabrda:
This PR makes us require RHEL 9.4 as a minimum version.

Additionally, it removes the exposed hostPort of the TUF server which shouldn't be exposed otherwise than through nginx.


SequeI (approved):
/lgtm

PR #149, opened 2025-02-20
Fixing health probe in Trillian-MySQL

SequeI:
During my testing of the restore procedure for ansible, I thought my terrible code was causing output errors just to find out after 2 hours that it was the health probes in the manifest.

Fix: Remove a space before the password :crying_cat_face:


fghanmi (approved):
LGTM! Thanks!

PR #151, opened 2025-02-24
Fixing missing volume mounts in rekor-redis and redis-backfill

SequeI:
Missing mounts forced redis to create anonymous ephemeral volumes. Fixed this by adding the required mounts.


fghanmi (approved):
LGTM! Thanks!

PR #154, opened 2025-02-26
[SECURESIGN-1469] Enable providing custom trust root content.

bkabrda:
This PR will allow users to provide archive with custom content of the Trust Root as a configuration variable that contains base64-encoded tar.gz file.

Note: I will extend the user-provided Molecule scenario for this usecase in another PR once we can override all of Fulcio, CTLog, TSA and Rekor targets (currently custom keys for Rekor and CTLog are impossible to configure).

### DIFF
--- a/roles/tas_single_node/README.md
+++ b/roles/tas_single_node/README.md
@@ -44,6 +44,7 @@ Deploy the [RHTAS](https://docs.redhat.com/en/documentation/red_hat_trusted_arti
 | tas_single_node_podman | Configuration options for Podman. | dict of 'tas_single_node_podman' options |  |
 | tas_single_node_cockpit | Configuration options for Cockpit. | dict of 'tas_single_node_cockpit' options |  `{'enabled': False, 'user': {'create': False, 'username': 'cockpit-user'}}`  |
 | tas_single_node_podman_volume_create_extra_args | Additional arguments to pass to the `podman volume create` command. This can be used to specify extra options when creating Podman volumes. | str |  |
+| tas_single_node_trust_root | Configuration options for the Trust Root. | dict of 'tas_single_node_trust_root' options |  `{'full_archive': ''}`  |
 
 #### Options for main > tas_single_node_rekor_redis
 
@@ -267,6 +268,12 @@ Deploy the [RHTAS](https://docs.redhat.com/en/documentation/red_hat_trusted_arti
 | create | Whether or not to create the cockpit user. | bool | no |  |
 | username | Username for the cockpit user. | str | no |  |
 | password | Password for the cockpit user. | str | yes |  |
+
+#### Options for main > tas_single_node_trust_root
+
+|Option|Description|Type|Required|Default|
+|---|---|---|---|---|
+| full_archive | A compressed base64-encoded .tgz file of the trust root. This archive file has a single directory named `repository/`, containing the contents of the TUF repository, for example `1.root.json`. The default value is an empty string. Using the default value generates the trust root content, unless a trust root already exists. Specifying a `full_archive` string removes any earlier configured trust root content, and starts to use the new specified content. If changing this value back to an empty string after setting the `full_archive` option, then we continue serving the previous value for the trust root content. To reset the trust root content, you must remove all files in the volume associated with the `tuf-repository` pod. | str | no |  |
 
 ## Example Playbook
 
--- a/roles/tas_single_node/tasks/podman/tuf.yml
+++ b/roles/tas_single_node/tasks/podman/tuf.yml
@@ -39,13 +39,89 @@
       data:
         ctfe-pubkey: |
           {{ (remote_tuf_certificates.results | selectattr('source', 'equalto', tas_single_node_remote_ctlog_public_key.0) | list | first).content }}
-        fulcio-cert: |
+        fulcio_v1.crt.pem: |
           {{ (remote_tuf_certificates.results | selectattr('source', 'equalto', tas_single_node_remote_fulcio_root_ca) | list | first).content }}
         rekor-pubkey: |
           {{ rekor_public_key_result.stdout | b64encode }}
-        tsa-cert: |
+        tsa.certchain.pem: |
           {{ (remote_tuf_certificates.results | selectattr('source', 'equalto', tas_single_node_tsa_certificate_chain) | list | first).content }}
+        # If user switches back from providing to not providing the full_archive, we have to trigger
+        # the rerun of the init job somehow; we do this by triggering a change in this secret.
+        __change_detection: "{% if 'full_archive' in tas_single_node_trust_root %}{{ 1 | b64encode }}{% else %}{{ 0 | b64encode }}{% endif %}"
   register: secret_result
+
+- name: Create the TUF repository volume
+  ansible.builtin.command:
+    cmd: "podman volume create --ignore {{ tas_single_node_tuf_repository_volume_name }}"
+  changed_when: false
+
+- name: Set TUF repository content to the provided trusted root archive
+  when: '"full_archive" in tas_single_node_trust_root and tas_single_node_trust_root.full_archive != ""'
+  block:
+    - name: Mount the TUF repository volume
+      ansible.builtin.command:
+        cmd: "podman volume mount {{ tas_single_node_tuf_repository_volume_name }}"
+      register: volume_mount_result
+      changed_when: false
+
+    - name: Create the directory to extract provided archive to
+      ansible.builtin.file:
+        state: directory
+        dest: "{{ tas_single_node_tuf_repo_dir }}"
+        mode: "0700"
+
+    - name: Create the archive file
+      ansible.builtin.file:
+        state: touch
+        access_time: preserve
+        modification_time: preserve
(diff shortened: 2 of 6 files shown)


roles/tas_single_node/tasks/podman/tuf.yml
```diff
@@ -45,8 +45,82 @@
...
         tsa-cert: |
           {{ (remote_tuf_certificates.results | selectattr('source', 'equalto', tas_single_node_tsa_certificate_chain) | list | first).content }}
+        # If user switches back from providing to not providing the full_archive, we have to trigger
+        # the rerun of the init job somehow; we do this by triggering a change in this secret.
+        __change_detection: "{% if 'full_archive' in tas_single_node_trust_root %}{{ 1 | b64encode }}{% else %}{{ 0 | b64encode }}{% endif %}"
   register: secret_result
 
+- name: Create the TUF repository volume
+  ansible.builtin.command:
+    cmd: "podman volume create --ignore {{ tas_single_node_tuf_repository_volume_name }}"
```
fghanmi, line 55: @bkabrda is this same volume intended to be used in [tuf-init.j2 ](https://github.com/securesign/artifact-signer-ansible/blob/3ed0257585625207e390b1af789f2e0e5bdcda13/roles/tas_single_node/templates/manifests/tuf/tuf-init.j2#L56)?

SequeI (approved):
lgtm - Thanks !

roles/tas_single_node/tasks/podman/tuf.yml
bkabrda: @fghanmi yes, that's the volume; the init job doesn't get run if the user is providing the archive.

fghanmi: yes, right, this volume would be used in [tuf.j2](https://github.com/securesign/artifact-signer-ansible/blob/3ed0257585625207e390b1af789f2e0e5bdcda13/roles/tas_single_node/templates/manifests/tuf/tuf.j2#L68), but what about `persistentVolumeClaim` ? in tuf.j2, it'll consider the k8s volume not the podman's (I suppose they're different), or am I getting it wrong ?

bkabrda: @fghanmi good question - I just looked into this more closely [in the docs](https://docs.podman.io/en/stable/markdown/podman-kube-generate.1.html) and the way this works is that the `claimName` is actually the name of the podman volume to mount, while the `name` volume of the attribute is just something you reference in `volumeMounts`. The fact they're the same value is "coincidence", but doesn't really hurt.

Does that help?

fghanmi: Interesting! Thanks for sharing this doc!

fghanmi (approved):
LGTM! Thanks!

roles/tas_single_node/README.md
```diff
@@ -237,6 +238,12 @@ Deploy the [RHTAS](https://docs.redhat.com/en/documentation/red_hat_trusted_arti
 | username | Username for the cockpit user. | str | no |  |
 | password | Password for the cockpit user. | str | yes |  |
 
+#### Options for main > tas_single_node_trust_root
+
+|Option|Description|Type|Required|Default|
+|---|---|---|---|---|
+| full_archive | Base64-encoded gzipped tarball of the Trust Root. This must contain a single directory `repository/` with the contents of the TUF repository (e.g. `1.root.json`). Note: switching from providing to not providing this value will not reset the Trust Root content. In this case, all files in the `tuf-repository` podman volume must be removed manually to reset the Trust Root content. | str | no |  |
```
ritz303, line 245: I've rewritten some of this based on our tech writing standards.  Here is my suggestion for the description text:

``A compressed base64-encoded .tgz file of the Trust Root.
This archive file has a single directory `repository/` with the contents of the TUF repository, for example `1.root.json`.
Changing this option does not reset the Trust Root content.
To reset the Trust Root content, you must remove all files in the volume associated with the `tuf-repository` pod.``

In the first sentence, I'm not sure of the file name extension, so I choose .tgz, but feel free to replace this if it's .gz or .tar instead.

Let me know what you think.

bkabrda: @ritz303 thanks! One note:

The .tgz works well I think.

What I was trying to make clear with the "reset" is that if the user first provides this value through their playbook, but then change their configuration and they stop providing it, they shouldn't expect the trust root to change, they'll need to wipe it manually if they want it to get automatically regenerated.
If they provide the value and then they provide a different value, then the new value will be respected.

IOW, changing the option from one value to different value does change the trust root; changing it from one value to nothing will not touch it. Does that make sense?

ritz303: @bkabrda : Not sure I'm following.  The way you are describing this gives me the impression that this option is either enabled or disabled, but I don't think that is the case since this option is not expecting a boolean value.  Your explanation seems to be contradictory to me.  You say if the user provides a value for this option in their playbook, then they "stop providing" a value sometime later, then nothing is changed with the trust root.  Then in your second paragraph (starting with IOW), you say changing the value to a different value does change the root trust.  So, isn't that what the user is doing in their playbook, changing this value to whatever?  What does "stop providing" mean?  Does that mean they are setting this option to an empty value (null)?  Does it or does it not change the trust root?  I'm confused.  What's the default string value?  Null?  Do you have an example of what this value can be?  Maybe that will help me understand.

bkabrda: @ritz303 sorry for the confusion, let me try to explain it better:

* The default is empty string; in this case if the trust root doesn't exist yet, it gets generated; if it exists it's just served as-is.
* When the user provides the `full_archive`, we wipe out whatever has been there previously and use the content of the provided archive.
  * Afterwards, if the user provides a different `full_archive` (e.g. an updated one), we will wipe the previous one and serve the newly provided one.
  * However, if the user changes this option to empty string after previously providing a `full_archive`, it's not obvious what should happen. What I want to say is that in this case we will continue serving the previously provided `full_archive`, even if the option is set to empty string now.
    * If they want to get rid of it and start "fresh" with an autogenerated archive by the collection, they need to wipe the trust root manually.

Does that make more sense?

ritz303: @bkabrda : Thank you for clarifying things for me!  It was helpful.  I've updated my suggestion for the option description based on this better understanding.  Here's the suggestion:

``A compressed base64-encoded .tgz file of the trust root. This archive file has a single directory named  `repository/`, containing the contents of the TUF repository, for example `1.root.json`. The default value is an empty string. Using the default value generates the trust root content, unless a trust root already exists. Specifying a `full_archive` string removes any earlier configured trust root content, and starts to use the new specified content. If changing this value back to an empty string after setting the `full_archive` option, then we continue serving the previous value for the trust root content. To reset the trust root content, you must remove all files in the volume associated with the `tuf-repository` pod.``

Hopefully that better captures what you want to say regarding this option.  Let me know what you think.

roles/tas_single_node/README.md
```diff
@@ -268,6 +269,12 @@ Deploy the [RHTAS](https://docs.redhat.com/en/documentation/red_hat_trusted_arti
 | username | Username for the cockpit user. | str | no |  |
 | password | Password for the cockpit user. | str | yes |  |
 
+#### Options for main > tas_single_node_trust_root
+
+|Option|Description|Type|Required|Default|
+|---|---|---|---|---|
+| full_archive | A compressed base64-encoded .tgz file of the trust root. This archive file has a single directory named `repository/`, containing the contents of the TUF repository, for example `1.root.json`. The default value is an empty string. Using the default value generates the trust root content, unless a trust root already exists. Specifying a `full_archive` string removes any earlier configured trust root content, and starts to use the new specified content. If changing this value back to an empty string after setting the `full_archive` option, then we continue serving the previous value for the trust root content. To reset the trust root content, you must remove all files in the volume associated with the `tuf-repository` pod. | str | no |  |
```
bouskaJ, line 276: @[bkabrda@redhat.com](mailto:bkabrda@redhat.com) this is different from what we have in operator. We do not have an option to provide the full_archive TUF repository. The TUF configuration consumes keys with metadata instead. The key rotation is done on the live container. Can we take a similar approach to unify both the configuration?

bkabrda: @bouskaJ we discussed this with others and came to the conclusion that this is a more Ansible-y approach and is easier on users; we think that it is reasonable to diverge in this case. Going forward, the TUF auto updater feature is going to be configured similarly on both the operator and the collection (it will be mutually exclusive with this setting).

We think this is a more practical alternative to copying the data over manually (do note that this is something that the user is supposed to use to provide manually updated trust root; the initial trust root still gets generated automatically - same as for operator)

bouskaJ: Ok, it seems that this is not the solution taken in the first second. :-D I understand that we will need to differ in some cases. I just wanted to point that out because once we create different APIs, it will be hard to unify them.

bkabrda: I totally get your point and I do agree with you that we need to strive for consistency; however, we might see cases where it's beneficial to slightly diverge and I do believe this is one of them. I always welcome more opinions, so thanks a lot either way!

fghanmi (approved):
LGTM!

PR #157, opened 2025-02-27
Refactoring default vars file to move individual config options into their component dict objects.

SequeI:
We are now adding all config options to their own nested object dicts. This PR refactors leftovers that were not done that way.

I left the  `ctlog_ca_passphrase` for Tommy to do in his current ctlog PR as there is no ctlog object dict in main yet - Just easier that way and less rebasing

Created new `tas_single_node_rekor` dict obj to hold the few rekor vars.(For now, until rekor-sharding populates it further)


tommyd450 (approved):
This pr looks good to me so far. Im assuming it isnt finished just yet

bkabrda (approved):

PR #160, opened 2025-03-03
Adding backup/restore feature

SequeI:
Backup/restore 1.2.0 feature

### DIFF
--- a/roles/tas_single_node/tasks/main.yml
+++ b/roles/tas_single_node/tasks/main.yml
@@ -2,6 +2,31 @@
 - name: Importing default values
   ansible.builtin.import_tasks: set_defaults.yml
 
+- name: Run Manual Backup and Exit
+  when: tas_single_node_backup_restore.backup.force_run and
+        not tas_single_node_backup_restore.restore.enabled
+  block:
+    - name: Fail if passphrase is not provided
+      ansible.builtin.fail:
+        msg: "Passphrase must be set for the backup to complete successfully and securely."
+      when: tas_single_node_backup_restore.backup.passphrase == ''
+
+    - name: Populate service facts
+      ansible.builtin.service_facts:
+
+    - name: Inform the user if tas_backup service doesn't exist
+      ansible.builtin.fail:
+        msg: "The tas_backup service does not exist. Please ensure a successful collection run has completed without the force_run option."
+      when: "'tas_backup.service' not in services"
+
+    - name: Run Backup Oneshot Job
+      ansible.builtin.systemd:
+        name: tas_backup.service
+        state: started
+
+    - name: Exit after Backup
+      ansible.builtin.meta: end_host
+
 - name: Check configuration correctness
   ansible.builtin.include_tasks: configuration_check.yml
 
@@ -14,6 +39,11 @@
 
 - name: Create Podman Volumes
   ansible.builtin.include_tasks: volumes.yml
+
+- name: Run TAS System Restore
+  ansible.builtin.include_tasks: podman/backup_restore/restore.yml
+  when: tas_single_node_backup_restore.restore.enabled and
+        not tas_single_node_backup_restore.backup.force_run
 
 - name: Create cockpit-user
   ansible.builtin.user:
--- a/roles/tas_single_node/tasks/main.yml
+++ b/roles/tas_single_node/tasks/main.yml
@@ -72,3 +72,7 @@
 
 - name: Execute Podman Actions
   ansible.builtin.include_tasks: podman.yml
+
+- name: Configure/Deploy Backup
+  ansible.builtin.include_tasks: podman/backup_restore/backup.yml
+  when: tas_single_node_backup_restore.backup.enabled | bool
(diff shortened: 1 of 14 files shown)


roles/tas_single_node/tasks/main.yml
```diff
@@ -1,4 +1,21 @@
...
+  block:
+    - name: Run Backup Oneshot Job
+      ansible.builtin.systemd:
+        name: tas_backup.service
+        state: started
+
+    - name: Exit after Backup
+      ansible.builtin.meta: end_host
+
+- name: Run TAS System Restore
```
bkabrda, line 14: I think we planned for restore to be a different role? What made you rethink this?

OTOH I can see how this easily fits in here. However, I would probably run it as the *last* task - this will make sure that everything, e.g. podman volumes, is precreated and we can unpack the archived data to the proper locations.

This will require a big fat disclaimer that this is only supposed to be used on a single run and then disabled, so that users don't keep it enabled and don't restore on every Ansible run.

roles/tas_single_node/tasks/podman/restore.yml
```diff
@@ -0,0 +1,80 @@
...
+    mode: "0644"
+
+- name: Extract Basename from File
+  ansible.builtin.set_fact:
+    restore_file: "{{ tas_single_node_data_protection.restore.source | basename }}"
+
+- name: Push Encrypted Backup Tarball to Remote Server
+  ansible.builtin.copy:
+    src: "{{ tas_single_node_data_protection.restore.source }}.tar.gz.enc"
+    dest: "/root/{{ restore_file }}.tar.gz.enc"
```
bkabrda, line 14: `/root/` is home directory of root; I'd suggest doing this in a dedicated subdirectory of `tas_single_node_data_protection.backup.directory`? You could name it same as the archive (without the file extension) to ensure certain degree of uniqueness of the name, too.

roles/tas_single_node/tasks/podman/restore.yml
```diff
@@ -0,0 +1,80 @@
...
+  ansible.builtin.file:
+    path: "/root/{{ restore_file }}.tar.gz"
+    state: absent
+
+- name: Stop Podman Containers
+  ansible.builtin.shell: podman stop $(podman ps -aq)
+  register: podman_stop_result
+  changed_when: podman_stop_result.rc == 0 and podman_stop_result.stdout | length > 0
+
+- name: Restore Podman Volumes
```
bkabrda, line 51: Similar to my comment on the backup script, this will need to explicitly mount the podman volumes and copy their content to the mounted location.

bkabrda (changes requested):
Hi Aleksy, thanks for opening the PR! I think overall it looks good, but there is room for improvement - I left bunch of comments inline, so please take a look at them.

One additional note that I didn't leave inline is for starting/stopping containers on the node - we basically start/stop all of them, as we assume RHTAS is the only system in containers running on the managed node. I think this is a fair assumption, but it certainly deserves an explicit callout when we're documenting it (if it turns out to be a problem, we can improve this by only stopping/starting our containers in future versions).

roles/tas_single_node/tasks/main.yml
SequeI: Late reply, forgot about this - Yea, I initially thought the restore would be much more expanded but it's actually really simple so I don't think it requires a whole new role boilerplate that will just take up space. And as you said, it slots in really nicely as is

roles/tas_single_node/tasks/volumes.yml
```diff
@@ -1,13 +1,17 @@
...
-      - rekor-redis-storage
-      - redis-backfill-storage
-      - "{{ tas_single_node_tuf_repository_volume_name }}"
-      - "{{ tas_single_node_tuf_signing_keys_volume_name }}"
+    volume_list: >-
+      {{
+        (["trillian-mysql"] if tas_single_node_trillian.database_deploy else []) +
+        [
+          "rekor-server",
+          "rekor-redis-storage",
```
bkabrda, line 9: I think using external Redis is also an option ATM, isn't it? Meaning that we would also have to conditionalize this (`tas_single_node_rekor_redis.database_deploy` can be used for the condition)

roles/tas_single_node/templates/scripts/backup.sh.j2
```diff
@@ -0,0 +1,84 @@
...
+BACKUP_DIR="{{ tas_single_node_backup_restore.backup.directory }}"
+CONFIG_DIR="{{ tas_single_node_config_dir }}"
+PASSPHRASE="{{ tas_single_node_backup_restore.backup.passphrase }}"
+
+SERVICES=(
+{% if tas_single_node_trillian.database_deploy %}
+  "trillian-mysql.service"
+{% endif %}
+  "rekor.service"
+  "redis.service"
```
bkabrda, line 15: Same as for the Redis volume - let's not include this if we're using external redis.

bkabrda (approved):
LGTM. AFAICS all comments have been addressed and I see no further issues. Thanks for your work on this, approved!

PR #165, opened 2025-03-05
handle fulcio private key passphrase + cert attributes changes

fghanmi:


roles/tas_single_node/tasks/certificates.yml
```diff
@@ -78,6 +78,33 @@
...
+    - name: Verify if the passphrase matches the auto_generated private key
+      ansible.builtin.shell:
+        cmd: >-
+          openssl ec -in '{{ tas_single_node_remote_fulcio_auto_generated_private_key }}' -check \
+           -passin 'pass:{{ tas_single_node_fulcio.ca_passphrase }}' > /dev/null 2>&1 || echo "failed"
+      register: check_key_passphrase
+      changed_when: false
+      when: (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_fulcio_auto_generated_private_key) | list | length) != 0
+
+    - name: Clean-up old generated private key
```
bkabrda, line 100: To be safe, I'd actually prefer if we created a backup of this file. It's ~1kB, so that doesn't hurt in terms of size and we could actually prevent users from losing their private keys if they accidentally change config in a bad way (this was the QE feedback and I think I agree with it).

This can be done by using the `backup` option of `ansible.builtin.copy`, but I think this doesn't apply here, because we're not copying a file from place A to B, we need to remove it first and then create it by piping openssl output. So in this case, I'd probably vote for an explicit `ansible.builtin.copy` with `state: absent` and `dest` something like `{{ tas_single_node_remote_fulcio_auto_generated_private_key + '.' + ansible_date_time.iso8601_basic }}`

Let's do the same for the certificate, if we're regenerating it.

bkabrda (changes requested):
Hi Firas, thanks for submitting the PR. It looks really good, I just left one inline comment about backing up old generated files; additionally, I would propose that we modify the docs to explicitly state that changing these values regenerates the Fulcio certificate and that requires accompanying manual change in the trust root.

roles/tas_single_node/README.md
```diff
@@ -23,7 +23,7 @@ Deploy the [RHTAS](https://docs.redhat.com/en/documentation/red_hat_trusted_arti
...
-| tas_single_node_fulcio | Details on the certificate settings for Fulcio. Includes organizational details, the user-provided private key for signing the root certificate, and the user-provided root certificate itself. | dict of 'tas_single_node_fulcio' options |  `{'certificate': {'organization_name': '', 'organization_email': '', 'common_name': ''}, 'private_key': '', 'root_ca': '', 'trusted_ca': '', 'ca_passphrase': 'rhtas', 'ct_log_prefix': 'rhtasansible'}`  |
+| tas_single_node_fulcio | Details on the certificate settings for Fulcio. Includes organizational details, the user-provided private key for signing the root certificate, and the user-provided root certificate itself. **Note**: Updating any of the certificate attributes (such as `organization_name`, `organization_email`, or `common_name`) will regenerate the Fulcio certificate, which requires a corresponding manual update in the trust root. | dict of 'tas_single_node_fulcio' options |  `{'certificate': {'organization_name': '', 'organization_email': '', 'common_name': ''}, 'private_key': '', 'root_ca': '', 'trusted_ca': '', 'ca_passphrase': 'rhtas', 'ct_log_prefix': 'rhtasansible'}`  |
```
bkabrda, line 26: I think this also applies to changing the private key password, when the private key is being autogenerated, is that correct?

bkabrda (approved):
LGTM, thanks!

PR #166, opened 2025-03-05
[SECURESIGN-1984] Fix handling of default values in nested variables

jourdee-lab:
This pull request addresses the issue where the user-provided values for nested dictionaries in the **tas_single_node** role override the entire dictionary causing other values to become undefined.
The changes ensures that the default values are properly combined and executed properly.

-  Remove all such defaults from defaults/main.yml

- Add them with a slightly changed name (e.g. prefixed with underscore) to vars/main.yml

-  Add a new task file (imported from tasks/main.yml right at the start) that will include tasks that properly combine our and user-provided values, e.g.:

### DIFF
--- a/roles/tas_single_node/defaults/main.yml
+++ b/roles/tas_single_node/defaults/main.yml
@@ -26,57 +26,11 @@ tas_single_node_trillian:
     host: trillian-mysql-pod
     port: 3306
 
-tas_single_node_ingress_certificates:
-  fulcio:
-    certificate: ""
-    private_key: ""
-  rekor:
-    certificate: ""
-    private_key: ""
-  tuf:
-    certificate: ""
-    private_key: ""
-  tsa:
-    certificate: ""
-    private_key: ""
-  rekor-search:
-    certificate: ""
-    private_key: ""
-  cli-server:
-    certificate: ""
-    private_key: ""
-
-tas_single_node_fulcio:
-  certificate:
-    organization_name: ""
-    organization_email: ""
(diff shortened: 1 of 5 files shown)


tommyd450 (changes requested):
Hey Pr looks good to me, I do have an ask, ctlog-config-template.j2 hard codes alot of default values, could you remove the defaults and add them to the defaults configuration you have been creating and strip the unneccesary jinja if logic.

roles/tas_single_node/defaults/main.yml
```diff
@@ -5,97 +5,12 @@ tas_single_node_registry_password: ""
...
-    organization_name: ""
-    organization_email: ""
-    common_name: ""
-  private_key: ""
-  root_ca: ""
-  ca_passphrase: rhtas
-  trusted_ca: ""
-  ct_log_prefix: rhtasansible
+tas_single_node_rekor_public_key_retries: 5
+tas_single_node_rekor_public_key_delay: 10
```
bkabrda, line 9: Why are we re-adding these? I thought they were moved under `tas_single_node_rekor`

jourdee-lab: Thank you for spotting that, I will correct it right away

bkabrda (approved):
LGTM! @tommyd450 please review as well for the ctlog related bits.

tommyd450 (approved):
Once the Readme is fixed, you can go ahead and merge, LGTM

PR #173, opened 2025-03-12
Default var readme fixes

SequeI:


roles/tas_single_node/defaults/main.yml
```diff
@@ -86,6 +86,7 @@ tas_single_node_tsa:
   trusted_ca: ""
 
 tas_single_node_ctlog:
+  ca_passphrase: rhtas
```
bkabrda, line 89: I think this will conflict with https://github.com/securesign/artifact-signer-ansible/pull/166, can we wait until that's merged (because that's been open for quite a while now) and then update this part of this PR?

SequeI: Sounds good

bkabrda (approved):
LGTM, thanks!

PR #177, opened 2025-03-13
Fix Fulcio/TSA root cert generation

fghanmi:


SequeI (approved):
lgtm, thanks for the fix!

bkabrda (approved):

PR #179, opened 2025-03-13
fix fulcio user-provided private key (string) verification

fghanmi:


SequeI:
@fghanmi Could we not just move this check AFTER the user provided key has been copied over into it's destination, then use the destination filepath variable? Would only take pasting this task after it and should work without need to create tmp files and such. lmk what you think

fghanmi:
> @fghanmi Could we not just move this check AFTER the user provided key has been copied over into it's destination, then use the destination filepath variable? Would only take pasting this task after it and should work without need to create tmp files and such. lmk what you think

Yes, I agree, I've updated the PR (with something similar to your idea, juste moved `user-provided Fulcio private key` task before the verification task, just to avoid going thru the rest of the tasks given that the verification might fail), 
Thanks!

SequeI (approved):
lgtm, thanks Firas!

PR #186, opened 2025-03-18
[SECURESIGN-2056] fix missing required arguments: ct_log_prefix

fghanmi:


roles/tas_single_node/templates/manifests/fulcio/fulcio-server.j2
```diff
@@ -57,7 +57,7 @@ spec:
             - /var/run/fulcio-secrets/cert.pem
             - --fileca-key-passwd
             - "{{ tas_single_node_fulcio.ca_passphrase }}"
-            - "--ct-log-url=http://{{ tas_single_node_ctlog_pod }}-pod:{{ tas_single_node_ctlog_port_http }}/{{ tas_single_node_fulcio.ct_log_prefix }}"
+            - "--ct-log-url=http://{{ tas_single_node_ctlog_pod }}-pod:{{ tas_single_node_ctlog_port_http }}/{{ tas_single_node_fulcio.ct_log_prefix | default(_tas_single_node_fulcio.ct_log_prefix) }}"
```
bkabrda, line 60: Is this necessary? The way we initialize the variables should make sure that if the value is unspecified, it will default to our default in `vars/main.yml`.

fghanmi: Actually, I wasn't aware of this change ! but it's great that we have it! 
Thanks!

bkabrda (approved):
LGTM, thanks!

PR #191, opened 2025-03-19
Running molecule tests from Mac

petrpinkas:
Updated `Containerfile` works fine for me from Mac. Test from Linux would be appreciated.

For `podman inspect` I used the similar approach as the one used two steps later. Hopefully it should provide the same output ...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,6 @@ context/
 *+00:00.json
 *.swp
 .vscode/
+.idea/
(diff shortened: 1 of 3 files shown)


bkabrda (approved):

SequeI (approved):

fghanmi (approved):

SequeI:
Verified on Linux, works as it should thanks!

PR #192, opened 2025-03-20
Add explicit dependency on the ansible.posix collection

bkabrda:
We need this dependency for the backup/restore machinery as well as for enabling custom trust root upload. As the minimum dep version I chose 1.5.4 - I tested with it and it seems fine, the newer versions seem to preserve backwards compability for the tasks we're using.


fghanmi (approved):
LGTM!

PR #195, opened 2025-03-21
molecule converge -> test 

SequeI:
Changed converge -> test
Fixed one task to pass idempotence checks

Should be all good now, esp for finding more issues and bugs

### DIFF
--- a/roles/tas_single_node/tasks/certificates.yml
+++ b/roles/tas_single_node/tasks/certificates.yml
@@ -169,6 +169,7 @@
             old_root_content != tas_single_node_fulcio.root_ca
             and (tas_single_node_fulcio.private_key != ""
               or (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_fulcio_auto_generated_private_key) | list | length) == 0)
+          changed_when: false
 
     - name: Create Fulcio private key
       # NOTE: As of cosign 2.3.0, only RSA with > 2048 B pkey size or ECC with prime256v1 (==secp256r1) are supported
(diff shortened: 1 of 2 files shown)


lance (approved):

roles/tas_single_node/tasks/certificates.yml
```diff
@@ -169,6 +169,7 @@
             old_root_content != tas_single_node_fulcio.root_ca
             and (tas_single_node_fulcio.private_key != ""
               or (certs_dir_files.files | selectattr('path', 'equalto', tas_single_node_remote_fulcio_auto_generated_private_key) | list | length) == 0)
+          changed_when: false
```
bkabrda, line 172: This `changed_when` shouldn't really be possible IMO. Assuming the `when` logic works fine, this task will only trigger if a certain config change happened, but rerunning again after that shouldn't trigger it again, should it?

roles/tas_single_node/tasks/certificates.yml
```diff
@@ -166,10 +166,9 @@
             remote_src: true
             mode: '0600'
           when: >
-            old_root_content != tas_single_node_fulcio.root_ca
+            old_root_content.content | b64decode != tas_single_node_fulcio.root_ca
```
fghanmi, line 169: @SequeI 
could you please do the same for https://github.com/securesign/artifact-signer-ansible/pull/195/commits/f9e9fd8071bd1cc5306ba8fa95a12e4517830fb9#diff-cf4cfdda66d5c440fedfb86a1ae26c54527bc09b5754f0b88546130774ccc7edR149 ? 

(better to have this change in the same PR) 
Thanks!

fghanmi (approved):
LGTM! 
Thanks!

PR #201, opened 2025-03-26
use separate storage volume options

fghanmi:


roles/tas_single_node/tasks/volumes.yml
```diff
@@ -23,7 +23,7 @@
     label: "{{ item }}"
 
 - name: Create volume if it does not exist
-  ansible.builtin.command: podman volume create {{ tas_single_node_podman_volume_create_extra_args }} {{ item }}
+  ansible.builtin.command: podman volume create {{ lookup('vars', 'tas_single_node_' + item | replace('-', '_') + '_create_args', default='') }} {{ item }}
```
bkabrda, line 26: How about making the configuration value a dict like this:
```
tas_single_node_podman_volume_create_extra_args:
  tuf_repository: ""
  tuf_signing_keys: ""
  # and so on
```

This would group the values together nicely and would also allow you to do something like `tas_single_node_podman_volume_create_extra_args[item]` here which I think is nicer.

bkabrda (approved):
LGTM, thanks!

PR #205, opened 2025-03-27
add missing attributes of ctlog sharding_config

fghanmi:
Attributes added based on: https://github.com/securesign/certificate-transparency-go/blob/main/trillian/ctfe/configpb/config.pb.go#L250

### DIFF
--- a/roles/tas_single_node/meta/argument_specs.yml
+++ b/roles/tas_single_node/meta/argument_specs.yml
@@ -437,7 +437,7 @@ argument_specs:
                 required: false
                 version_added: "1.2.0"
               public_key:
-                description: "Public key used for verification"
+                description: "Public key used for verification (must be in DER format)"
                 type: "str"
                 required: false
                 version_added: "1.2.0"
--- a/roles/tas_single_node/meta/argument_specs.yml
+++ b/roles/tas_single_node/meta/argument_specs.yml
@@ -437,7 +437,7 @@ argument_specs:
                 required: false
                 version_added: "1.2.0"
               public_key:
-                description: "Public key used for verification (must be in DER format)"
+                description: "Public key used for verification (must be in DER format)."
                 type: "str"
                 required: false
                 version_added: "1.2.0"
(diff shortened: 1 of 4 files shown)


roles/tas_single_node/templates/configs/ctlog-config-template.j2
```diff
@@ -10,6 +10,53 @@ log_configs:{
...
                 password: "{{ log.password }}"
                 }}
-        ext_key_usages:"CodeSigning"  log_backend_name:"trillian"}
+        ext_key_usages: "CodeSigning"  
+        log_backend_name: "trillian"
+        {% if log.get('not_after_limit') -%}
+        not_after_limit: {{ log.not_after_limit | to_json | regex_replace('"([^"]+)":', '\\1:') | regex_replace('"(\d+)"', '\\1') }}
+        {% endif -%}
+        {% if log.get('public_key') -%}
+        public_key: {der:"{{ log.public_key }}"}
```
bkabrda, line 19: I would assume most users would provide the key in PEM format; we should either call the DER format out explicitly in the docs (or do the conversion ourselves).

osmman: Only DER format is supported in this configuration

bkabrda (approved):

roles/tas_single_node/meta/argument_specs.yml
```diff
@@ -435,6 +431,95 @@ argument_specs:
                 type: "str"
                 required: false
                 version_added: "1.2.0"
+              override_handler_prefix:
+                description: "Overrides the handler prefix for this log"
+                type: "str"
+                required: false
+                version_added: "1.2.0"
+              public_key:
+                description: "Public key used for verification (must be in DER format)"
```
tommyd450, line 440: Can we get the descriptions to be ended with a period. "."

Other than that it lgtm.

tommyd450 (approved):

PR #222, opened 2025-04-04
Pruning backup resources if set to false

SequeI:


bkabrda:
Hi Aleksy, thanks for the PR. Do you think it would be possible to use a different approach? It would be really nice to modify the current backup.yml and set the state of the resources there (service enabled/disabled, file present/absent) based on whether the user enabled the feature or not. Does that make sense?

roles/tas_single_node/tasks/podman/backup_restore/backup.yml
```diff
@@ -1,40 +1,92 @@
...
+      loop:
+        - "{{ tas_single_node_systemd_directory }}/tas_backup.service"
+        - /usr/local/bin/tas_backup.sh
+        - "{{ tas_single_node_systemd_directory }}/tas_backup_timer.timer"
+
+    - name: Reload Systemd Daemon
+      ansible.builtin.systemd:
+        daemon_reload: true
+
+- name: Run Manual Backup and Exit
```
bkabrda, line 69: By moving this block here, does it mean we're now allowing to run a backup on the first run, e.g. on a yet-to-be configured node?

SequeI: No, in this block there is a check for the tas_backup.service which will fail if it does not exist, and let the user know he needs to do a full run first to deploy the service

bkabrda: But that service is created in the first block of this file, isn't it?

SequeI: Oh. You are totally right and I am a dummy 😄

PR #225, opened 2025-04-08
Fixing systemd template formatting for container teardown and broken deployment restoration

SequeI:
1. Formatting for stopping containers and tearing down pod was faulty - Fixed. (Pod stopped first to get rid of runtime artifacts incase of backup, worked but second part for teardown did not because systemd commands are not ran in shell)

2. Incase of faulty deployment, added a restore.enabled check when restarting podman service since initially it would not, leaving behand broken pod remnants. (Complete manual teardown to ensure a mimic of a 'fresh node' documented in notes for Aron)


bkabrda (approved):
Nice catches, thanks!

PR #232, opened 2025-04-09
Rekor Pub Key Utilise Id

tommyd450:


SequeI (approved):

roles/tas_single_node/tasks/configuration_check.yml
```diff
@@ -22,6 +22,15 @@
           ==  (((tas_single_node_rekor.private_keys | list) + (tas_single_node_rekor.public_keys| list ))  | map(attribute='id') | list | length )
     msg: "Custom keys declared within 'tas_single_node_rekor' must have unique ID's."
 
+- name: Check
+  when: (tas_single_node_rekor.sharding_config | length > 0)
+  ansible.builtin.assert:
+    that: (key_var.pem_pub_key != "" and ( "-----BEGIN PUBLIC KEY-----" is in key_var.pem_pub_key ))
```
bkabrda, line 28: I think because we only want to recognize this value as the "ID", we should remove the "BEGIN PUBLIC KEY" part, right?

bkabrda (approved):
LGTM, thanks!

roles/tas_single_node/tasks/configuration_check.yml
tommyd450: :ballot_box_with_check:

bkabrda (approved):
LGTM, thanks for doing all the changes!

PR #305, opened 2025-07-11
Fix AAP CI

fghanmi:
This is due to change in `infra.aap_configuration` and latest version.
Related Issues: 
https://github.com/redhat-cop/infra.aap_configuration/issues/1102
https://github.com/redhat-cop/infra.aap_configuration/issues/1068

## Summary by Sourcery

Fix the Molecule AAP CI converge playbook by replacing the deprecated controller_license role with a custom task sequence and align subscription handling with the latest infra.aap_configuration changes.

Bug Fixes:
- Swap inclusion of infra.aap_configuration.controller_license role for an include_tasks call to a new subscription.yml file
- Add filters for subscription lookup instead of forcing license install
- Introduce subscription.yml to fetch subscriptions and apply controller license via ansible.controller modules

### DIFF
--- a/molecule/aap-test/converge.yml
+++ b/molecule/aap-test/converge.yml
@@ -164,8 +164,7 @@
       delegate_to: localhost
       block:
         - name: Include role to add subscription to controller
-          ansible.builtin.include_role:
-            name: infra.aap_configuration.controller_license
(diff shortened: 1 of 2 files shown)


SequeI (approved):
lgtm, thank you!

PR #332, opened 2025-08-04
fix: configure mysql client connection pool limits

osmman:
Refs: SECURESIGN-2650


osmman:
> @osmman I noticed that you've added the Rekor MySQL connection pool arguments in the [operator PR ](https://github.com/securesign/secure-sign-operator/pull/1255):
> 
> * `--search_index.mysql.conn_max_idletime`
> * `--search_index.mysql.conn_max_lifetime`
> * `--search_index.mysql.max_idle_connections`
> * `--search_index.mysql.max_open_connections`
> 
> however, I don't see these in this ansible PR, should we add these to the ansible side as well?

Hi, this project doesn't currently have a feature to configure the search index, which is why it hasn't been added yet. For now, the only option is to use the Redis implementation.

PR #426, opened 2025-11-24
test: Disable GH actions testing in favour of Konflux

bouskaJ:
## Summary by Sourcery

Align CI to Konflux by relying on Tekton pipelines and removing redundant GitHub Actions workflows and actions.

CI:
- Expand Tekton pull request and push pipelines to trigger when Molecule tests or testing requirements change.
- Remove GitHub Actions workflows and composite actions related to molecule testing, collection build, upgrades, and go e2e testing in favor of the Tekton-based pipeline.


SequeI:
imo we should leave the code itself, just disable it incase Konflux stops working or is down, so we can atleast test. wdyt?

bouskaJ:
> imo we should leave the code itself, just disable it incase Konflux stops working or is down, so we can atleast test. wdyt?

Yes, I was thinking about that and I decided that we can restore pipelines from git history but you are right that we may enable/disable on the Konflux outage and be more flexible with the source-code present. I am going to close this PR and disable workflows manually in UI.

PR #555, opened 2026-01-27
fix: ansible e2e

bouskaJ:


fghanmi:
@bouskaJ could you please rebase so AAP test gets updated/fixed ?

PR #790, opened 2026-06-26
Update release-version to 1.4.2

petrpinkas:


tommyd450 (changes requested):
We need to change the versions in the Dockerfile, galaxy.yml and the readme.

